PR #29, opened 2025-10-02
Kh spi

kentnhong:
Add SPI driver class with factory functions. Just need to fix switch logic for the setter functions of the cr settings. Also need to implement SPI read, write, and transfer.


TJMalaska (changes requested):
I think this is un updated lmk when you wnat a review

PR #33, opened 2025-10-03
PH spi

pbho48985219:


common/drivers/platform/stm32l4/st_spi.h
```diff
@@ -0,0 +1,85 @@
...
+};
+
+enum class SpiEnable : std::uint8_t{
+    Enable = 0,
+    Disable
+};
+class HwSpi
+{ //pins should already be configured
+public: 
+    explicit HwSpi(SPI_TypeDef* spi, Pin mosi, Pin miso, Pin clk); //passing SPI1, SPI2, etc
```
TJMalaska, line 53: Change the typedef name from "spi" to base_addr to keep everything consistent. I also dont think we should pass in the gpio for mosi etc... But im not really sure what you mean with "pin". What type is that?

common/drivers/platform/stm32l4/st_spi.h
```diff
@@ -0,0 +1,85 @@
...
+    //constructor declaration for a class: HwSpi ^
+    bool spi_init();
```
TJMalaska, line 55: When we have no arguments, it's practice to put "void" inside the args

common/drivers/platform/stm32l4/st_spi.cc
```diff
@@ -0,0 +1,81 @@
...
+/*
+NAMESPACE:
+- groups related code (classes, functions,variables) avoiding conflicts
+- different than parent classes: Organizing groups while parent classes are for data structure 
+- doesn't affect code, just groups them to organize (name wrapper)
+
+WRAPPERS:
+- makes it easier to use, but doesn't change functionality 
+*/
+void HwSpi::spi_init(){ //from HwSpi
```
TJMalaska, line 25: There's not much more review I can do until is seen how you init.

TJMalaska (changes requested):
Overall great progress, much better than I expected :). You have the right ideas and are definitely learning the concepts. Seems about halfway done. Your next goal is to fix all of my comments, finish init, and make the SpiModule implementations to test.

PR #38, opened 2025-11-08
bs-imu

pasinponsww:


common/drivers/platform/bno055/bno055_imu.h
```diff
@@ -0,0 +1,219 @@
...
+*/
+
+#pragma once
+#include <cstdint>
+#include "imu.h"
+#include "stm32l4xx.h"
+#include "st_i2c.h"
+
+namespace LBR {
+    class Bno055 : public Imu {
```
nfan17: doc strings

common/drivers/platform/bno055/bno055_imu.h
```diff
@@ -0,0 +1,219 @@
...
+* @author Bex Saw
+* @date 2025-10-22
+* @note This file contains the interface for the BNO055 IMU sensor, including initialization, reading, and calibration functions.
+*/
+
+#pragma once
+#include <cstdint>
+#include "imu.h"
+#include "stm32l4xx.h"
+#include "st_i2c.h"
```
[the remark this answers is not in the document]
nfan17: note: think about what hardware targets it should be possible to run a piece of code on, if you find yourself including libraries that are more specific than the supposed scope of usability then you're probably including something wrong

common/drivers/platform/bno055/bno055_imu.h
```diff
@@ -0,0 +1,219 @@
...
+    class Bno055 : public Imu {
+    public:
+        explicit Bno055(LBR::Stml4::HwI2c& i2c_, uint8_t addr = 0x28);
+
+        // Core IMU interface methods
+        void Init() override;
+        void Update() override;
+
+        // Sensor data retrieval methods
+        void GetAcceleration(float& ax, float& ay, float& az) override;
```
nfan17: define data type for 3D vector, so you can use that instead of individually defined ax, ay, az, etc

thinking about it from a user perspective:
```
Vec3 accel, gyro, euler;
bno.get_accel(accel);
bno.get_gyro(gyro);
bno.get_euler(euler);
```
looks a lot better than
```
float ax, ay, az, gx, gy, gz, roll, pitch, yaw;
bno.get_accel(ax, ay, az);
.... etc
```

common/drivers/platform/bno055/bno055_imu.h
```diff
@@ -0,0 +1,219 @@
...
+
+        // I2C interface and device address
+        LBR::Stml4::StI2c& i2c_;
+        uint8_t address_;
+        uint8_t accelData_[6] = {0}; // X, Y, Z accelerometer data
+        uint8_t gyroData_[6] = {0};  // X, Y, Z gyroscope data
+        float euler_[3] = {0.0f, 0.0f, 0.0f}; // Roll, Pitch, Yaw
+
+        //Default I2C address for BNO055
+        static constexpr uint8_t BNO055_I2C_ADDR = 0x28;
```
nfan17: this changes depending on addr config so have defs for both maybe?

common/drivers/platform/bno055/bno055_imu.h
```diff
@@ -0,0 +1,219 @@
+/**
+* @file bno055_imu.h
+* @brief BNO055 IMU sensor interface
+* @author Bex Saw
+* @date 2025-10-22
+* @note This file contains the interface for the BNO055 IMU sensor, including initialization, reading, and calibration functions.
+*/
+
```
nfan17: I would advise against storing the values and only updating all of them synchronously, it should be possible to call each function and immediately access real hw to pull the individual value

for example, if someone wants to use just accel they now have to deal with the overhead of polling everything

app/imu_test/bsp_sensors/imu_app_bsp.h
```diff
@@ -0,0 +1,16 @@
...
+*/
+
+#pragma once
+
+#include "st_i2c.h"
+#include "bno055_imu.h"
+#include "stm32l4xx.h"
+
+
+LBR::Bno055& BSP_Init_IMU(I2C_TypeDef* i2c_instance, GPIO_TypeDef* gpio_instance);
```
nfan17: where board?

app/imu_test/main.cc
```diff
@@ -0,0 +1,84 @@
...
+[FlightComputer]
+    ↓ (printf / UART or something else not sure yet)
+[USB/Serial Virtual COM Port]
+    ↓ (Real-time data packets)
+[Ayush’s UI Application]
+*/
+
+int main(void) {
+    // 1. Initialize Board & IMU
+    auto& imu = BSP_Init_IMU(I2C1, GPIOB);
```
nfan17: this will change eventually anyways but a) poor assumption that both gpio pins use the same port, and also there shouldn't be hardware specific details in your main

app/imu_test/bsp_sensors/imu_app_bsp.cc
```diff
@@ -0,0 +1,44 @@
...
+* @date 2025-10-24
+*/
+
+#include "imu_app_bsp.h"
+#include "st_i2c_module.h"
+#include "delay.h"
+#include "st_gpio.h"
+
+LBR::Bno055& BSP_Init_IMU(I2c_TypeDef* i2c_instance, GPIO_TypeDef* gpio_instance) {
+    (void)gpio_instance;
```
nfan17: NTO USED ok so delete it entirely

nfan17 (commented):
not bad! overall try to think more about what it would actually be like to consume your own APIs (writing functions isn't just a matter of exposing functionality, it's also abt making intuitive sets of calls that are easy to use in an application). Also, a little more attention to where you're using hardware specific code will help you understand when to decouple things like hardware and application logic

common/drivers/bus/i2c.cc
```diff
@@ -0,0 +1,101 @@
...
+
+#include "i2c.h"
+
+namespace LBR
+{
+bool I2c::mem_read(std::span<uint8_t> data, const uint8_t reg_addr,
+                   uint8_t dev_addr)
+{
+    // Creating span for register address
+    const uint8_t reg[1] = {reg_addr};
```
nfan17: no raw c arrays, use a std::array. Then you can pass in without creating a span at all

std::array<uint8_t, 1> reg {reg_addr};
ret = ret && burst_write(reg, dev_addr);

nfan17 (commented):
@yshiblanco sorry didn't see ur PR but these 2 comments r for u

TJMalaska (commented):
I need to still review the bno functions

app/imu_test/bsp_sensors/imu_app_bsp.cc
```diff
@@ -0,0 +1,122 @@
+/**
+* @file imu_app_bsp.cc
```
TJMalaska: Don't need doc strings for .cc files.

common/drivers/platform/bno055/bno055_imu.h
```diff
@@ -0,0 +1,151 @@
...
+ */
+class Bno055
+{
+
+public:
+    /**
+     * @enum Mode
+     * @brief IMU operating modes
+     */
+    enum class Mode : uint8_t
```
TJMalaska: We don't have to use an enum class here, this is one of the times we can use a normal enum. A enum class is used so there isn't implicit type conversion but since these values are basically pasted into a registers we want them to be implicitly converted to unit8s without static casting. Nit pick.

common/drivers/platform/bno055/bno055_imu.cc
```diff
@@ -0,0 +1,219 @@
...
+    out.quat.y = combine(buf[idx + 4], buf[idx + 5]) / QUAT_SCALE;
+    out.quat.z = combine(buf[idx + 6], buf[idx + 7]) / QUAT_SCALE;
+
+    return true;
+}
+
+bool Bno055::calibrate(uint8_t& value)
+{
+    // CALIB_STAT register address is 0x35
+    return i2c_.mem_read(std::span<uint8_t>(&value, 1), (uint8_t)0x35,
```
TJMalaska: Here and functions below you are doing c style casts which shouldn't be done in c++. Can you also make a static constexper in the function that holds the value for the address just to make it more readable instead of the comment.

common/drivers/platform/bno055/bno055_imu.h
```diff
@@ -0,0 +1,151 @@
...
+ * @note This file contains the interface for the BNO055 IMU sensor, including initialization, reading, and calibration functions.
+ */
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include "../../../core/math/imu_math.h"
+#include "../../bus/i2c.h"
+#include "../../utils/delay.h"
```
TJMalaska: We dont need to to relative imports!

common/drivers/platform/bno055/bno055_imu.cc
```diff
@@ -0,0 +1,219 @@
...
+ * 
+ */
+void Bno055::deinit()
+{
+    // Put device into deep suspend to save power
+    set_mode(Mode::CONFIG);
+    // Per BNO055 datasheet, delay after switching to CONFIG mode.
+    LBR::Utils::DelayMs(10);
+    uint8_t pwr_mode = 0x02;  // SUSPEND
+    i2c_.mem_write(std::span<const uint8_t>(&pwr_mode, 1), (uint8_t)0x3E,
```
TJMalaska: You dont need to cast this if you make pwr mode a std array.

TJMalaska (approved):

PR #46, opened 2025-11-20
yb-i2c-hotfix

yshiblanco:


common/drivers/bus/i2c.h
```diff
@@ -22,19 +22,25 @@ class I2c
      * @brief Read data from external device
      * 
      * @param data block of memory to read data into from the bus
+     * @param reg_addr data register of external device to read from
      * @param dev_addr address of target device
      * @return true if successful, false otherwise
      */
-    virtual bool mem_read(std::span<uint8_t> data, uint8_t dev_addr) = 0;
+    virtual bool mem_read(std::span<uint8_t> data,
+                          std::span<const uint8_t> reg_addr,
```
pasinponsww: Add the burst_read and burst_write to the i2c.h as a pure virtual (=0) and then and mem_read and mem_write dont have to be pure virtual b/c they will be implement using the burst_read and burst_write. You can put the implementation on the i2c.h or i2c.cc, it can make a (mem_read & mem_write) generic :) - Nick & TJ

pasinponsww (changes requested):
Make changes :)

common/drivers/bus/i2c.h
yshiblanco: fixed

common/drivers/bus/i2c.h
```diff
@@ -22,19 +22,25 @@ class I2c
...
      * @brief Writes data to external device
      * 
      * @param data block of memory storing data to write into the bus
+     * @param reg_addr data register of external device to write to
      * @param dev_addr address of target device
      * @return true if successful, false otherwise
      */
-    virtual bool mem_write(std::span<const uint8_t> data, uint8_t dev_addr) = 0;
+    virtual bool mem_write(std::span<const uint8_t> data,
+                           std::span<const uint8_t> reg_addr,
```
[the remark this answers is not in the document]
yshiblanco: fixed

common/drivers/bus/i2c.cc
```diff
@@ -0,0 +1,99 @@
...
+    buf[0] = reg_addr;
+    for (size_t i = 0; i < data.size(); i++)
+    {
+        buf[1 + i] = data[i];
+    }
+
+    std::span<const uint8_t> data_buf{buf, 1 + data.size()};
+
+    // Data transfer
+    bool ret = true;
```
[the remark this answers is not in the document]
yshiblanco: done

common/drivers/platform/stm32l4/st_i2c.h
TJMalaska: Don't need to include span if it's included a layer up.

common/drivers/platform/stm32l4/st_i2c.h
```diff
@@ -26,6 +26,19 @@ struct StI2cParams
     uint32_t timingr;
 };
 
+enum TransferMode : uint8_t
```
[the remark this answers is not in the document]
TJMalaska: Also for the MODE it makes sense to use a enum class for type safety instead of a regular enum. The only reason I can think of is when doing bit manipulation, so the read write enum is fine.

common/drivers/platform/stm32l4/st_i2c.cc
```diff
@@ -60,23 +121,40 @@ bool HwI2c::burst_read(std::span<uint8_t> data, uint8_t dev_addr)
...
+
+    // Configure according to mode
+    switch (mode)
+    {
+        case GENERATE_STARTSTOP:
+            _base_addr->CR2 |= I2C_CR2_AUTOEND;
+            break;
+        case GENERATE_START_ONLY:
+            break;
+        case GENERATE_STOP_ONLY:
```
TJMalaska: What does the stop only do? I understand startstop and start only. Does it just skip sending the dev address? This interface is kinda of risky because we can have an attempted i2c transfer with no address event sent. Not sure how to feel about it.

TJMalaska: Why not just call detect_stop Im confused what this is used for.

common/drivers/platform/stm32l4/st_i2c.cc
```diff
@@ -37,9 +38,69 @@ bool HwI2c::init()
...
 
-bool HwI2c::burst_read(std::span<uint8_t> data, uint8_t dev_addr)
+bool HwI2c::mem_read(std::span<uint8_t> data, const uint8_t reg_addr,
+                     uint8_t dev_addr)
 {
+    bool ret{true};
+
+    ret = ret && initiate_transfer(dev_addr, GENERATE_START_ONLY, WRITE, 1);
+    std::array<uint8_t, 1> reg{reg_addr};
+    std::span<const uint8_t> register_addr{reg};
```
TJMalaska: You don't need to create a span from an array to pass it into burst_write. C++ implicitly can cast a std::array into a std::span so you can skip that. This applies everywhere you are doing this.

common/drivers/platform/stm32l4/st_i2c.cc
```diff
@@ -127,55 +161,149 @@ bool HwI2c::initiate_transfer(uint8_t dev_addr, uint8_t mode, uint8_t dir,
...
+    {
+        // TODO: replace while with timeout
+    }
+    _base_addr->ICR |= I2C_ICR_STOPCF;
 
     return true;
 }
 
-bool HwI2c::burst_read(std::span<uint8_t> data)
+bool HwI2c::burst_read(std::span<uint8_t> data, uint8_t dev_addr)
```
TJMalaska: FUCK! we should probably make the burst write and read an actual burst write and read instead of raw. Im such a bot.

yshiblanco: i changed name to just read and write

yshiblanco: or wait what did u want me to do lol

pasinponsww (approved):
LGTM!

TJMalaska (approved):

PR #58, opened 2025-12-26
kh-w25q

kentnhong:
First PR for the W25Q flash chip driver. Everything compiles correctly. Have not tested on actual flash chip yet.


common/core/periph/flash_chip/w25q.h
```diff
@@ -0,0 +1,57 @@
...
+#include <cstring>
+#include <span>
+#include "gpio_cs.h"
+
+#include "spi.h"
+
+namespace LBR
+{
+
+enum class StatusWrite : uint8_t
```
nfan17: ```LBR::StatusWrite::XYZ``` and ```LBR::StatusRead:::XYZ``` might be too generic of a name to place in global LBR namespace, I'd consider moving it under the W25q class

common/core/periph/flash_chip/w25q.cc
```diff
@@ -0,0 +1,543 @@
...
+    // Wait for current writes or erases to finish
+    while (this->BusyCheck())
+    {
+    }
+
+    cs.ChipSelectEnable();
+    std::array<uint8_t, 4> txbuf{0x3Du, static_cast<uint8_t>(block_addr >> 16),
+                                 static_cast<uint8_t>(block_addr >> 8),
+                                 static_cast<uint8_t>(block_addr)};
+    spi.Transfer(txbuf, block_lock_status);
```
nfan17: this spi.Transfer may be named a little too vaguely, my assumption was that it would send and receive simultaneously (which after reading the implementation I now know it is not and is sequential). also I THINK you can use enhanced for loops for both your TX and RX loops in the implementation:

```
for (const auto & byte: tx_data)
    {
        ...

        // Write next tx byte to start clocking
        *(volatile uint8_t*)&instance->DR = byte;
        
        ...
    }
```

and

```
    for (auto & byte: rx_data)
    {
        ...
        // Read data from RX buffer into rx_data
        byte = *(volatile uint8_t*)&instance->DR;
    }
```

I always use these when I can to avoid stupid errors writing the for loop portion

common/core/periph/flash_chip/w25q.cc
```diff
@@ -0,0 +1,543 @@
...
+
+    return true;
+}
+
+/**
+ * @brief Check if an individual block is locked or not
+ * 
+ * @param block 
+ * @param block_lock_status 
+ * @return true Block is currently locked
```
nfan17: nit: ```@return True if block is currently locked, false otherwise```

one return statement makes things a bit more concise

common/core/periph/flash_chip/w25q.cc
```diff
@@ -0,0 +1,543 @@
...
+    cs.ChipSelectDisable();
+
+    return true;
+}
+
+/**
+ * @brief Check if an individual block is locked or not
+ * 
+ * @param block 
+ * @param block_lock_status
```
nfan17: why allocate this outside of the function? also does this even work, or does it make a copy since it's not a std::array reference? it's easier to make a separate helper function that ONLY parses the response and returns a boolean, while this does the actual transaction. Given the current flow, there's no way to know if this transaction failed and if the return value is based on true success or whatever value is in the array you passed into it.

common/core/periph/flash_chip/w25q.cc
```diff
@@ -0,0 +1,543 @@
+#include "w25q.h"
+
+/**
+ * @brief Construct a new LBR::W25q::W25q object
+ * 
+ * @param spi_ 
+ */
+LBR::W25q::W25q(Spi& spi_, GpioChipSelect& cs_) : spi(spi_), cs(cs_)
```
nfan17: prefer bracket intializers for everything:

```spi{spi_}, cs{cs_}```

File: common/core/periph/flash_chip/w25q.h (hunk 1 above)
Comment (kentnhong) [reply common/core/periph/flash_chip/w25q.h]: done

File: common/core/periph/flash_chip/w25q.h [hunk 6]
```diff
@@ -0,0 +1,57 @@
...
+    STATUS_REGISTER_1 = 0x05u,
+    STATUS_REGISTER_2 = 0x35u,
+    STATUS_REGISTER_3 = 0x15u
+};
+
+class W25q
+{
+public:
+    W25q(Spi& spi_, GpioChipSelect& cs_);
+
```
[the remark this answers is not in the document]
TJMalaska: i been telling this to people man...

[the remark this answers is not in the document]
kentnhong: done

common/core/periph/flash_chip/w25q.cc
kentnhong: done

common/core/periph/flash_chip/w25q.cc
```diff
@@ -0,0 +1,543 @@
...
+ * @return true Block was successfully unlocked
+ * @return false Block was already unlocked
+ */
+bool LBR::W25q::BlockUnlock(uint8_t block)
+{
+    std::array<uint8_t, 1> block_lock_status;
+
+    // Calculate address of which block to check
+    uint32_t addr = 0x00;
+    addr += (static_cast<uint32_t>(block) * 0x010000u);
```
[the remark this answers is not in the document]
kentnhong: done

pasinponsww (dismissed):
LGTM!

PR #60, opened 2026-01-01
bs-pps

pasinponsww:
**PPS Core Abstraction Layer**
- note: (IMU, GPIO, Motor) and system logic (state machine, safety, motion control)
* add necessary files
* skeleton for state machine


app/pps_test/pps.h
```diff
@@ -0,0 +1,73 @@
...
+   
+    PpsState state_ = PpsState::Idle;
+        // IMU data fields (quaternion, etc.)
+        //...
+        // PWM control fields
+        //...
+
+        // GPIO/limit switch state
+        bool limit_switch_min_ = false;
+        bool limit_switch_max_ = false;
```
TJMalaska: What do the min and max limit switched mean?

app/pps_test/pps.cc
```diff
@@ -0,0 +1,66 @@
...
+* @file pps.cc
+* @brief PPS module implementation
+* @note This module manages the positioning system state machine.
+* @author Bex Saw
+* @date 2025/12/31
+*/
+
+
+#include "pps.h"
+#include "bsp/motor_if.h"
```
TJMalaska: All of these includes should just come in the pps.h so the motor if can be included there and will be linked to this automatically.

app/pps_test/pps.cc
```diff
@@ -0,0 +1,66 @@
...
+
+namespace LBR {
+
+Pps::Pps() {
+    state_ = PpsState::Idle;
+    // Init IMU, PWM, etc. if needed (not sure yet)
```
TJMalaska: We should not be initing and hardware in a constructor ever because there's no way to check success. We will set up hardware with a function somewhere else (we can talk if you need a solution to that).

app/pps_test/pps.cc
```diff
@@ -0,0 +1,66 @@
...
+    // limit_switch_min_ = readLmsMin();
+    // limit_switch_max_ = readLmsMax();
+}
+
+void Pps::update() {
+    // Read sensors
+    // auto [roll, pitch, yaw] = getOrientation(); from IMU
```
TJMalaska: Im open to other thoughts but since the IMU will be polling for flash data there's no reason to update separately here, Also i2c drivers is NOT thread safe. A possible solution for this is updating the sensor values in one location and have other components only read.

app/pps_test/pps.cc
```diff
@@ -0,0 +1,66 @@
...
+    // limit_switch_min_ = readLmsMin(); from GPIO
+    // limit_switch_max_ = readLmsMax(); from GPIO
```
TJMalaska: Im fine with the gpio limit switches being updated in here.

app/pps_test/pps.cc
```diff
@@ -0,0 +1,66 @@
...
+            // Limit switch: Monitor only
+            if (shouldHome()) state_ = PpsState::Homing;
+            break;
+        case PpsState::Homing:
+            // Homing: Move motor toward home position
+            // PWM: Drive motor in home direction
+            // Limit switch: Check home (min) switch, stop when triggered
+            if (limit_switch_min_) state_ = PpsState::MovingToTarget;
+            break;
+        case PpsState::MovingToTarget:
```
TJMalaska: We need to be alot more clear with these functions. Is moving to target a degree from the starting location? a distance? what units?

app/pps_test/motor_support/motor_if.h
```diff
@@ -0,0 +1,39 @@
+#pragma once
```
TJMalaska: Small note for the motor interface. This will be generic for a DC motor! This means that none of this should be specific to the PPS. The drilling app will probably use this as well.

app/pps_test/motor_support/motor_if.h
```diff
@@ -0,0 +1,39 @@
...
+ * TODO: Implement PWM logic for rotation.
+ */
+void motorTarget();
+
+/**
+ * @brief Move the motor in the retract direction (retract mechanism).
+ * @note Used during the Retract state to drive the mechanism until the retracted limit switch is reached.
+ * TODO: Implement PWM logic for retraction.
+ */
+void motorRetract();
```
TJMalaska: All of these functions are PPS specific so they should not be in the DC motor class. The DC motor class will be generic and will also be used for the drilling app. All of these function though are useful for the pps so maybe we can move them to another file like pps_helpers.

app/pps_test/main.cc
```diff
@@ -0,0 +1,24 @@
+/**
```
TJMalaska: We won't need docstrings for a main, its just redundant and bloat.

app/pps_test/pps.h
```diff
@@ -0,0 +1,68 @@
...
+public:
+    PpsState getState() const;
+    Pps();
+    void fetchImuData(const LBR::Bno055Data& imu_data); // Update IMU data only
+    void update(); // State machine update, no IMU arg
+
+private:
+    PpsState state_ = PpsState::Idle;
+
+     bool limit_switch_max = false; // Deployed position
```
TJMalaska: Brace initialize these. Its common practice for modern C++ even though they get optimized to about the same thing.

app/pps_test/pps.h
```diff
@@ -0,0 +1,68 @@
...
+     bool limit_switch_min = false; // Retracted position
+
+    /**
+     * @brief Read the current state of the limit switch (PA3).
+     * @return true if the switch is active, false otherwise.
+     */
+     static bool readLimitSwitchMax();
```
TJMalaska: None of these functions gotta be static?

app/pps_test/pps.h
```diff
@@ -0,0 +1,68 @@
...
+     /**
+      * @brief Read the current state of the min limit switch (e.g. PA4, retracted).
+      * @return true if the switch is active, false otherwise.
+      */
+     static bool readLimitSwitchMin();
+    /**
+     * @brief Check if deploy sequence should start.
+     * @return true if deployment should begin, false otherwise.
+     * @note Used to determine when to transition from Idle to Deploy state.
+     */
```
TJMalaska: Both min and max can be one single function. Return 0 if at the start and 1 if we're extended.

TJMalaska (commented):
good work

Image

PR #70, opened 2026-01-14
yb-servo

### LINKED ISSUE #71: Servo Motor Driver
Implement an interface for controlling servo motors

yshiblanco:


pasinponsww:
rebase + change range PWM

pasinponsww:
We need to fix the servo to test the angle rotation based on the pwm current drive the servo motor

**[Link](https://www.amazon.com/KEAcvise-Coreless-Stainless-Waterproof-Robotic-270/dp/B0FGPJQ6BD/ref=sr_1_1_sspa?adgrpid=)**

.vscode/launch.json
```diff
@@ -151,6 +151,7 @@
       }
     },
     {
+<<<<<<< HEAD
```
pasinponsww, line 154: I think you accidently ignore this changes. What I would suggest is copy the latest launch.json from the main and then add your test at the bottom. Because some of this line sneak into the .json it will mess up the file process! please take a look!

.vscode/launch.json
```diff
@@ -178,6 +179,35 @@
         "monitor reset"
       ],
       "showDevDebugOutput": true
+=======
```
pasinponsww, line 182: Yeah usually it will show you the current changes vs incoming changes you have to look it and make comparison which one is the most up to date. This is good attempt.

.vscode/launch.json
```diff
@@ -178,6 +179,35 @@
...
+                "${workspaceRoot}",
+            ],
+            "preRestartCommands": [
+                "add-symbol-file build/stm32l476/app/pwm_test/pwm_test",
+                "load",
+                "enable breakpoint",
+                "monitor reset"
+            ],
+            "showDevDebugOutput": true,
+>>>>>>> e4a3cee (small change to app; fixed .json)
```
pasinponsww, line 210: Yeah it is sneak into here so take a look!

app/servo_test/bsp_h723/h723_board.cc
```diff
@@ -0,0 +1,68 @@
...
+#include "board.h"
+#include "st_gpio.h"
+#include "st_pwm.h"
+
+// -----------------------------------------------------------------------
+// Tune these after physical calibration
+// -----------------------------------------------------------------------
+static constexpr uint16_t MIN_PULSE_WIDTH = 500;   // microseconds at 0° (0.5ms)
+static constexpr uint16_t MAX_PULSE_WIDTH = 2500;  // microseconds at MAX_ANGLE (2.5ms)
+static constexpr uint16_t MAX_ANGLE       = 270;   // degrees (270° servo)
```
pasinponsww, line 19: You want to define all of the static constexpr in this format

`static constexpr uint16_t MIN_PULSE_WIDTH{500};
`
b/c in modern C++, brace initialization {}, or list initialization, is generally preferred over the assignment = syntax because it forbids narrowing conversions and provides a uniform syntax across all types. However, for simple fundamental types like uint16_t, there is zero difference in the compiled machine code but still you should do it for consistency!

app/servo_test/bsp_h723/h723_board.cc
```diff
@@ -0,0 +1,68 @@
...
+
+namespace LBR
+{
+
+uint32_t kPclkFreq = 64000000;  // confirm with Kent if clock speed differs
```
pasinponsww, line 24: so yes this will be different. Since we going to use the HSE with different operating frequency after PLL. But for now keep it this way! as it is the default freq!

app/servo_test/bsp_h723/h723_board.cc
```diff
@@ -0,0 +1,68 @@
...
+                                     pwm_settings, kPclkFreq};
+
+Stmh7::HwGpio pwm_output(pwm_output_params);
+Stmh7::HwPwm  pwm(pwm_params);
+
+// Servo motor
+const ServoMotorParams servo_motor_params{pwm, MIN_PULSE_WIDTH,
+                                          MAX_PULSE_WIDTH, MAX_ANGLE};
+ServoMotor servo_motor(servo_motor_params);
+
```
pasinponsww, line 48: So I see all of the instance here, servo object, pwm, gpio. But I do not see any sysclk instance here. You should add the StSysclock instance in here which is still consider the hardware layer in this board support package!

app/servo_test/board.h
```diff
@@ -0,0 +1,22 @@
...
+ * @date 01/28/2026
+ */
+
+#pragma once
+#include "servo_motor.h"
+
+namespace LBR
+{
+
+struct Board
```
pasinponsww, line 14: I have question as of where is the 

Gpio& pwm_out;
...
and the output instance. Can you answer me if you know where its define? usually we put our instance in here board.h like Gpio& I2c& Usart& and so on 

But in your h723_board.cc has:
 ret = ret && pwm_output.init();
 ret = ret && pwm.init();

Which Im assume it should be 

Pwm& pwm
Gpio& pwm_output 


give me a good line of reasoning if you know anything

app/servo_test/main.cc
```diff
@@ -0,0 +1,27 @@
...
+    {
+        angle += 30;
+        if (angle > 270)
+        {
+            angle = 0;
+        }
+
+        hw.servo_motor.set_angle(angle);
+
+        for (volatile size_t i = 0; i < 200000; i++);
```
pasinponsww, line 23: we going to use the delay_ms() now no more the busy wait delay because busywait completely freezes the CPU and wastes massive power, while a hardware timer counts independently in the background

pasinponsww (changes requested):
Please make some changes! it's not bad but some of the question and refactor!

PR #72, opened 2026-01-14
tm_sysclock

TJMalaska:
Goal of this PR is to enable higher sys clock frequencies.
Has been tested on GPIO
Needs to be tested on:
SPI
I2C


nfan17 (approved):
I say get rid of clock generic interface but otherwise lgtm!

PR #75, opened 2026-01-15
bs-drv8245

pasinponsww:
motor driver 

references:
image

image


common/drivers/platform/drv8245/drv8245.cc
```diff
@@ -0,0 +1,36 @@
+#include "drv8245.h"
+
+namespace LBR {
+
+Drv8245::Drv8245(Gpio& dir, Pwm& pwm, Gpio& drv_z, Gpio& sleep, Gpio& fault)
+    : dir_(dir), pwm_(pwm), drv_z_(drv_z), sleep_(sleep), fault_(fault)
```
kentnhong: Use brace constructor initialization over parentheses constructor initialization for stricter type checks. It makes it so your code won't compile if for example dir is a different type than dir_.

`
Drv8245::Drv8245(Gpio& dir, Pwm& pwm, Gpio& drv_z, Gpio& sleep, Gpio& fault)
    : dir_{dir}, pwm_{pwm}, drv_z_{drv_z}, sleep_{sleep}, fault_{fault} {}
`

common/drivers/platform/drv8245/drv8245.cc
```diff
@@ -0,0 +1,36 @@
...
+}
+
+void Drv8245::setPwm(Direction dir, uint16_t pwm_value) {
+    dir_.set(dir == Direction::Forward ? true : false);
+    pwm_.set_duty_cycle(static_cast<uint8_t>(pwm_value / 10)); // Convert 0-1000 to 0-100%
+}
+
+void Drv8245::enableCoast() {
+    drv_z_.set(false); // Disable driver outputs (Hi-Z)
+}
```
kentnhong: This is a contuation of my comment on returning bool for your methods. For example, since we know the .set() function from gpio.h returns a bool, we can use this to our advantage to check if enableCoast() suceeded.

`return drv_z_.set(false);`

In this case, I don't think TJ implemented false logic for his .set() function so it can only return true if all the code in .set() is executed.

common/drivers/platform/drv8245/drv8245.cc
```diff
@@ -0,0 +1,36 @@
...
+
+void Drv8245::init() {
+    drv_z_.set(true); // Enable driver
+    sleep_.set(true);  // Wake up driver
+}
+
+void Drv8245::setPwm(Direction dir, uint16_t pwm_value) {
+    dir_.set(dir == Direction::Forward ? true : false);
+    pwm_.set_duty_cycle(static_cast<uint8_t>(pwm_value / 10)); // Convert 0-1000 to 0-100%
+}
```
TJMalaska: This function isn't necessary in my opinion, we can just use the interface that yshi made for the PWM.

pasinponsww:
Closed-PR new sub branch

PR #107, opened 2026-03-07
jc-i2c

pasinponsww:


app/imu_test/bsp_h723/h723_board.cc
```diff
@@ -0,0 +1,82 @@
+#include <stdint.h>
+#include "board.h"
+#include "delay.h"
+#include "st_gpio.h"
+#include "st_i2c.h"
+#include "stm32h7xx_hal.h"  // for the hal_init
+// #include "st_sys_clock.h"
```
pasinponsww: Keep it as is for now. @kentnhong will create an H7-specific sysclk for you to use soon.

app/imu_test/bsp_h723/h723_board.cc
```diff
@@ -0,0 +1,82 @@
...
+Stmh7::HwGpio rst(rst_params);
+
+// Construct IMU driver using generic I2c interface
+Bno055 imu(static_cast<LBR::I2c&>(i2c), Bno055::ADDR_PRIMARY);
+Board board{.imu = imu};
+
+bool bsp_init()
+{
+    // Initialize system clock first
+    HAL_Init();
```
pasinponsww: Add the same comment above this one too (remove when sysclk is done)

common/drivers/platform/stm32h7/st_i2c.cc
```diff
@@ -0,0 +1,372 @@
...
+
+bool HwI2c::mem_read(std::span<uint8_t> data, const uint8_t reg_addr,
+                     uint8_t dev_addr)
+{
+    /* Check if ready */
+    if (!busFree())
+    {
+        return false;
+    }
+
```
pasinponsww: Add a check to confirm init() was called. If not initialized, return false.

pasinponsww (commented):
Good Job! Some minor changes such as structuring, consistency, doc string (for intellisense when u hover you can see what's the function does), @kentnhong still need to review, the logic in st_i2c.cc doesn't look sus to me. Please make a changes accordingly! Thx U

PR #108, opened 2026-03-07
sl-usart 

sethlunas:
usart for h7 specific

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -318,6 +318,6 @@
         "monitor reset"
       ],
       "showDevDebugOutput": true
-    },
+    }
   ]
 }


app/usart_app/BSP_H723/usart_app_bsp.cc
```diff
@@ -0,0 +1,74 @@
...
+{
+
+StGpioSettings tx_config = {GpioMode::ALT_FUNC, GpioOtype::PUSH_PULL,
+                            GpioOspeed::LOW, GpioPupd::NO_PULL, 0x7};
+
+StGpioSettings rx_config = {GpioMode::ALT_FUNC, GpioOtype::PUSH_PULL,
+                            GpioOspeed::LOW, GpioPupd::NO_PULL, 0x7};
+
+StGpioParams tx_params = {tx_config, (uint8_t)8, GPIOD};
+StGpioParams rx_params = {rx_config, (uint8_t)9, GPIOD};
```
pasinponsww: 1. Using namespace this way
```
namespace stmh7
{
 // content 
}
```
is better than using namespace LBR::stmh7 b/c name collisions and make it un clear where functions or types come from 

When selecting a pins for the StGpioSettings you want to keep it consistent with the other bsp, good idea, and it still initialized but we want to create a 

```
const Stml4::StGpioParams tx_params{tx_config, 8, GPIOD}; 
... so on
```

 is safer and clearer than an unqualified b/c it enforces explicit context

PR #109, opened 2026-03-10
ap-spi

alexanderpulido:
spi.h for the STM32H723xx, provide  interface and configuration types for a hardware SPI master driver. Declares HwSpi class inheriting from Spi, with read, write, seq_transfer, and init to be implemented in the accompanying .cpp.
image
image
image
image

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -342,6 +342,53 @@
         "monitor reset"
       ],
       "showDevDebugOutput": true
+    },
+    {
+      "cwd": "${workspaceFolder}",
+      "executable": "${workspaceRoot}/build/stm32h723/app/spi_test/spi_test",
+      "name": "DEBUG SPI TEST H7",
+      "request": "launch",
+      "type": "cortex-debug",
+      "servertype": "openocd",
+      "device": "STM32H723",
+      "svdFile": "${workspaceRoot}/mcu_support/stm32/h7xx/STM32H723.svd",
+      "searchDir": ["${workspaceRoot}/mcu_support/stm32"],
+      "runToMain": true,
+      "configFiles": [
+        "${workspaceRoot}/mcu_support/stm32/stlink-dap.cfg",
+        "${workspaceRoot}/mcu_support/stm32/h7xx/stm32h7x.cfg"
+      ],
+      "preRestartCommands": [
+        "add-symbol-file build/stm32h723/app/spi_test/spi_test",
+        "load",
+        "enable breakpoint",
+        "monitor reset"
+      ],
+      "showDevDebugOutput": true
+    },
+    {
+      "cwd": "${workspaceFolder}",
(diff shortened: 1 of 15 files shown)


common/drivers/platform/stm32h7/st_spi.h
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * @brief SPI status codes returned to indicate the result of an operation.
+ *
+ */
+enum class SpiStatus : uint8_t
+{
+    OK = 0,
+    read_ERR,
+    write_ERR,
```
pasinponsww:
```
READ_ERR
WRITE_ERR
```

keep it consistent keep it all caps

common/drivers/platform/stm32h7/st_spi.h
```diff
@@ -0,0 +1,175 @@
...
+class HwSpi : public Spi
+{
+public:
+    /**
+     * @brief Construct a new HwSpi object.
+     *
+     * @param instance_ Pointer to the SPI peripheral to use (e.g. SPI1, SPI2)
+     * @param settings_ Struct containing the desired SPI configuration
+     */
+    explicit HwSpi(SPI_TypeDef* instance_, StSpiSettings& settings_);
```
pasinponsww: Requested naming cleanup: change `instance_` to `instance` and `settings_` to `settings` less time typing and its cleaner that way

common/drivers/platform/stm32h7/st_spi.cc
```diff
@@ -0,0 +1,224 @@
...
+
+namespace LBR
+{
+namespace Stmh7
+{
+
+HwSpi::HwSpi(SPI_TypeDef* instance_, StSpiSettings& settings_)
+    : instance{instance_}, settings{settings_}
+{
+}
```
pasinponsww: For simplicity, once `settings_` and `instance_` are change to what I ask you in st_spi.h, the constructor becomes much cleaner.

pasinponsww (changes requested):
LGTM! But we need to test if it's work!

pasinponsww:
Testing on the oscope and can merge later

pasinponsww:
Need to verify and test and it's done

.vscode/launch.json
```diff
@@ -345,6 +345,36 @@
     },
     {
       "cwd": "${workspaceFolder}",
+      "executable": "${workspaceRoot}/build/stm32h723/app/spi_test/spi_test",
```
pasinponsww, line 347: We need to validate this at the oscilloscope level as well for testing signal. Since the default system clock is 64 MHz, we’ll also need to adjust the peripheral clock values and step down the system clock frequency accordingly.

pasinponsww (approved):
LGTM! since it's verified in person with my already I will merge this!

PR #121, opened 2026-04-22
jc-airbrake-app

ioseph01:
Needs preset angles and target altitudes, altitude formula, etc..


app/airbrake_test/control/airbrake.cc
```diff
@@ -0,0 +1,132 @@
+#include "airbrake.h"
+#include <cstdint>
+
```
pasinponsww: As I reference you about how you construct your application, lets put this in **common/core/control**

app/airbrake_test/control/airbrake.cc
```diff
@@ -0,0 +1,132 @@
...
+ *          /
+ *         /    Target 2
+ *        /     
+ *       /      Target 1
+ * .____/
+ * 
+ */
+constexpr float kTargetAlt1;
+constexpr float kTargetAlt2;
+constexpr float kTargetAlt3;
```
pasinponsww: Put this in an enum class **(or config struct)** so we can switch between altitude profiles cleanly instead of having loose constexpr values.

app/airbrake_test/control/airbrake.cc
```diff
@@ -0,0 +1,132 @@
...
+ *      |\/
+ *      |~\
+ *      |  \
+ *      |   \   
+ *      |    \
+ * 
+ */
+constexpr uint8_t kTargetAng1;
+constexpr uint8_t kTargetAng2;
+constexpr uint8_t kTargetAng3;
```
pasinponsww: This one as well in the struct or enum class

app/airbrake_test/control/airbrake.h
```diff
@@ -0,0 +1,89 @@
...
+{
+
+
+    enum class AirbrakeState
+{
+    WAITING,
+    ASCENT,
+    APOGEE,
+    DEATH, // temp val, if something has gone terrible wrong
+};
```
pasinponsww: Commend or Docstring for this too, we want to make it easy to read.

app/airbrake_test/control/airbrake.h
```diff
@@ -0,0 +1,89 @@
...
+* 3. Fault tolerance (Safety)
+* - Dual Trigger deploy IF (G-Force > X AND Altitude > Y)
+* - Servo Watchdog cuts powers to servo motor if it draws too much current
+* NOTE : Trust IMU for direction, Baro for height
+* 
+* 4. Input :
+* - CAN Input for LAUNCH and TARGET APOGEE from Telem board
+* - CAN Output send Servo angle and System Health to Telem board
+* - PWM Output drives the servo motor
+*/
```
pasinponsww: Let's throw this in the /doc folder under the same control folder. And do a .md so we can have it on the side.

app/airbrake_test/control/airbrake.h
```diff
@@ -0,0 +1,89 @@
...
+    WAITING,
+    ASCENT,
+    APOGEE,
+    DEATH, // temp val, if something has gone terrible wrong
+};
+
+class AirBrake
+{
+public:
+    AirBrake(motor);
```
pasinponsww: `AirBrake(motor);` is missing a type, so this won’t compile. Also, since this constructor takes one dependency, it should probably be `explicit` to avoid unintended implicit construction.

Suggested:
```cpp
explicit AirBrake(ServoMotor& motor);

File: app/airbrake_test/control/airbrake.h [hunk 7]
```diff
@@ -0,0 +1,89 @@
...
+
+    AirbrakeState getState() const;
+
+    void fetchImuData(const LBR::Bno055Data& data); // Fetch IMU Data
+
+    void fetchPressData(const float data);  // Fetch Pressure data (hPA)
+    
+    void fetchTempData(const float data);  // Fetch Temperature data (Celsius)
```
pasinponsww: **const** on pass-by-value primitive types does not help the caller and just clutters the interface. The state machine constantly taking data from one state to another.

app/airbrake_test/control/airbrake.h
```diff
@@ -0,0 +1,89 @@
...
+
+    void update();  // State machine update
+
+private:
+
+    float calcAltitude() const;
```
pasinponsww: No need when we already get the certain number threshold calculated by the Aero/MechE Team

app/airbrake_test/control/airbrake.h
```diff
@@ -0,0 +1,89 @@
...
+
+    AirbrakeState state = AirbrakeState::WAITING;
+    float pressure, temperature;
+    ServoMotor motor;
```
pasinponsww: ServoMotor& motor;

app/airbrake_test/control/airbrake.cc
```diff
@@ -0,0 +1,132 @@
...
+}
+
+void AirBrake::update() {
+  /**
+   * Get CAN INPUT FOR :
+   * TARGET_TELEM
+   * LAUNCH_SIGNAL
+   */
+
+  float altitude = calcAltitude();
```
pasinponsww: no more since we will get a set const vals from the AE or MechE

app/airbrake_test/control/airbrake.cc
```diff
@@ -0,0 +1,132 @@
...
+  bool launch_signal; // temp placeholder
+
+  switch (state)
+  {
+  case AirbrakeState::WAITING:
+    /* Get CAN input and check if ascent */
```
pasinponsww: Remove CAN. The airbrake runs autonomously based on input thresholds, with no external commands. Requires thorough testing to ensure the flaps do not trigger unintentionally.

app/airbrake_test/control/airbrake.cc
```diff
@@ -0,0 +1,132 @@
...
+    if (launch_signal /* LAUNCH SIGNAL */ ) {
+      state = AirbrakeState::ASCENT;
+    }
+    else { break; }
+
+  case AirbrakeState::ASCENT:
+    /* Set pwm to idk when at apogee */
```
pasinponsww: Set duty cycle to 100% to fully deploy the flaps (bounded by the servo’s angle limits).

app/airbrake_test/control/airbrake.cc
```diff
@@ -0,0 +1,132 @@
...
+     * so need to do anything with PWM?
+     */
+
+    // uint8_t be compared with floats
+    if (kTargetAlt1 <= altitude && altitude < kTargetAlt2) {
+      motor.set_angle(kTargetAng1);
+    }
+    else if (kTargetAlt2 <= altitude && altitude < kTargetAlt3) {
+      motor.set_angle(kTargetAng2);
+    }
```
pasinponsww: This is smart good job!

app/airbrake_test/control/airbrake.cc
```diff
@@ -0,0 +1,132 @@
...
+    else if (kTargetAlt3 <= altitude) {
+      state = AirbrakeState::APOGEE;
+    }
+    break;
+    
+  case AirbrakeState::APOGEE:
+    /** 
+     * Placing this here will repeatedly set angle to 90 (or whatever val).
+     * Not sure if motor will preserve the angle automatically or not.
```
pasinponsww: The servo compares that to its current position and drives the motor until it matches, then keeps applying small corrections to hold that angle.

app/airbrake_test/control/airbrake.cc
```diff
@@ -0,0 +1,132 @@
...
+    motor.set_angle(90); 
+    break;
+  default:
+    break;
+  }
+
+  /* 
+    Do we still need to calculate current velocity to compare with
+    target velocity?
+  */
```
pasinponsww: Right now, the state machine only compares air pressure values, but the airbrake decision should also use IMU data.

The IMU already gives us acceleration and orientation, so we should include checks like vertical acceleration, pitch/yaw/roll, and current orientation before deploying. Air pressure can tell us altitude/velocity trend, but IMU data helps confirm the rocket’s actual motion and attitude so the flaps do not deploy at the wrong time.

app/airbrake_test/control/airbrake.cc
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * Send CAN OUTPUT FOR :
+   * Angle
+   * Health?
+   */
```
pasinponsww: No board-to-board data needs to be sent back that’s already handled by the **/network** (my current task) via usart_pipe for real-time raw data transfer. I’ll just pull the state from your airbrake app; that’s all I need, and I’m working on it now.

pasinponsww (changes requested):
Looks decent to me. This is really simple state machine don't stress it.

ioseph01:
Right now, it's still simple state machine.

common/core/control/docs/info.md
```diff
@@ -0,0 +1,35 @@
...
+ 
+                                             Peak Altitude
+                                                   _
+                        Target Apogee ----->      / \      
+                                                 /   \
+                                                /     \
+                                               /       \     
+                                              /         \        
+         Waiting Starts        Ascent --->   /           \
+ _______________v___________________________/             \
```
pasinponsww: Image

Cleanup your README when you have a chance, it's not neccesary but it just break at display

common/core/control/docs/info.md
```diff
@@ -0,0 +1,35 @@
...
+   current velocity to compare with target velocity
+
+ 3. Fault tolerance (Safety)
+ - Dual Trigger deploy IF (G-Force > X AND Altitude > Y)
+ - Servo Watchdog cuts powers to servo motor if it draws too much current
+ NOTE : Trust IMU for direction, Baro for height
+ 
+ 4. Input :
+ - CAN Input for LAUNCH and TARGET APOGEE from Telem board
+ - CAN Output send Servo angle and System Health to Telem board
```
pasinponsww: I hope this is not **"CAN" like Control Area Network.** 

We don't send any command, we boot the hardware of the rocket and it set a baseline like altitude at sea level, etc. and then launch and your entire FSW (statemachine/airbrake_app) will handle all of the job. 

but if it's mean "can" as "able" then this make sense but "CAN Input for LAUNCH and TARGET APOGEE" I hope that means change the constant of at which altitude to launch, etc. 

If you can clarify this to me this will be great !

common/core/control/airbrake_constants.h
```diff
@@ -0,0 +1,22 @@
...
+namespace LBR
+{
+
+namespace AirbrakeConstants
+{
+constexpr int kMotorBurnTime = 6;      // In seconds
+constexpr float kMotorBurnAlt = 3250;  // in ft
+constexpr float kTargetApogee = 7500;  // in ft
+constexpr int kRetractAngle = 0;       // In degrees
+constexpr int kMaxAngle = 30;          // In degrees
```
pasinponsww: compiles fine, but silently truncates 6.8 to 6 (Dangerous for mission math!)

define it in {} format like this: 
```
constexpr int kMotorBurnTime{6};       // In seconds
constexpr float kMotorBurnAlt{3250.0f}; // In ft
constexpr float kTargetApogee{7500.0f}; // In ft
constexpr int kRetractAngle{0};        // In degrees
constexpr int kMaxAngle{30};           // In degrees
```

reason:  it stops silent data loss by forcing the compiler to reject dangerous type mismatches.  

maybe make it float for for MotorBurnTime too, we might need to factor in the **ms** as well like 6.7 for instance.

common/core/control/airbrake.h
```diff
@@ -0,0 +1,62 @@
...
+enum class AirbrakeState
+{
+    PRELAUNCH,                 // Not launched yet
+    LAUNCHED,                  // 6 second interval from launch
+    AIRBRAKES_DEPLOYING,       // Deploy airbrakes after 6 seconds
+    AIRBRAKES_FULLY_DEPLOYED,  // When at 30 degrees; not sure when to know though
+    AIRBRAKES_RETRACTING,      // Should do before target apogee
+    AIRBRAKES_FULLY_RETRACTED,  // Again, dont know how to know if retracted
+    RECOVERY_POPPED,            // Job finished
+};
```
pasinponsww: **enum class AirbrakeState : uint8_t** 
and then 
{
PRELAUNCH = 0,
...
}

The reason is: Strict Memory Control -> Standard un-scoped enum types default to int (typically 32-bits), which wastes RAM and Flash in embedded environments.

common/core/control/airbrake.h
```diff
@@ -0,0 +1,62 @@
...
+
+struct AirbrakeParams
+{
+    ServoMotor motor;
+    // Not sure what to put in here besides servo motor tbh.
+    // Maybe imu and baro, maybe not
```
pasinponsww: You don't need other objects to pass down into the struct. Just only ServoMotor or just feed it directly to the constructor but for simplicity just put in the struct for now.

common/core/control/airbrake.h
```diff
@@ -0,0 +1,62 @@
...
+};
+
+class AirBrake
+{
+public:
+    AirBrake(AirbrakeParams);
```
pasinponsww, line 37: explicit in front of the constructor ( prevents unintended implicit type conversions )

common/core/control/airbrake.h
```diff
@@ -0,0 +1,62 @@
...
+
+    AirbrakeState getState() const;
+
+    void fetchImuData(const LBR::Bno055Data& data);  // Fetch IMU Data
+
+    void fetchPressData(const float data);  // Fetch Pressure data (hPA)
+
+    void fetchTempData(const float data);  // Fetch Temperature data (Celsius)
+
+    void update();  // State machine update
```
pasinponsww: doc string with
```
/**
* @brief 
* @params
* @return etc. ykwim 
*/ 
```
For all function give it's meaningful meaning

common/core/control/airbrake/airbrake.cc
```diff
@@ -0,0 +1,147 @@
...
+    : motor{params_.motor}, state{AirbrakeState::PRELAUNCH}
+{
+}
+
+bool AirBrake::init()
+{
+    /**
+   * Maybe we could use the barometer to calculate altitude to set base_altitude.
+   * 
+   */
```
pasinponsww: To calculate a relative altitude baseline with the **BMP390**, the system must capture local atmospheric pressure immediately at boot. 

This init logic should be isolated inside a private helper function. As the project scales to incorporate complex mathematics, these algorithms I suggest throw in a new file (airbrake_utils.cc / .h) under the core/control/utils/ dir. 

Don't take it literally please contact the Mechanical, Tyler R. (EE) and see if this formula or anything online can help us get the baseline altitude at sea level. 

This might help: https://www.grc.nasa.gov/www/k-12/airplane/atmosmet.html (I did not look at it deeply yet but lmk)

Image

common/core/control/airbrake.cc
```diff
@@ -0,0 +1,147 @@
...
+
+    switch (state)
+    {
+        case AirbrakeState::PRELAUNCH:
+
+            if (launch_signal)
+            {
+                state = AirbrakeState::LAUNCHED;
+            }
+            break;  // <- This be removed? Nahhh
```
pasinponsww: no don't take it off it will break the statemachine lol

common/core/control/airbrake/airbrake.cc
```diff
@@ -0,0 +1,147 @@
...
+        case AirbrakeState::AIRBRAKES_FULLY_RETRACTED:
+            // Not sure what to do here
+            /**
+             * Maybe if acceleration or velocity downward, transition state
+             * 
+             * Or if receive signal for recovery popped but this not needed if so
+             */
+            state = AirbrakeState::RECOVERY_POPPED;
+            break;
+
```
pasinponsww, line 107: The current structure looks okay to me. Their going to be changes later in the future. but good skeleton ideas

app/airbrake_test/main.cc
```diff
@@ -0,0 +1,44 @@
...
+int main(int argc, char* argv[])
+{
+    bsp_init();
+    Board hw = get_board();
+
+    hw.imu.get_chip_id(chip_id);
+    hw.imu.get_opr_mode(opr_mode);
+    hw.imu.get_sys_status(sys_status);
+    hw.imu.run_post(self_test);
+    hw.imu.get_sys_error(sys_error);
```
pasinponsww: The imu specific like checking the chip_id, operating mode, etc. essentially we might just do a helper function that just check the sensor health at the start and send back to us thru uart debug header or serial comm maybe. Just be aware for now that we want to wrap this so it's not bloat main.

pasinponsww (changes requested):
Some styling changing, the logic of the airbrake looks good to me (I don't think we know until we test with the mock), we need to find out about the formula for the altitude baseline just so we have a good baseline for our baro operating point.

common/core/control/docs/info.md
ioseph01: Stale comment. Lot's of stale comments I have forgotten to remove.

app/airbrake_test/main.cc
ioseph01: Addressed but not fully resolved

common/core/control/airbrake/airbrake.cc
ioseph01: Using yes

ioseph01: Either read in init or maybe do it in prelaunch state?

pasinponsww: We need to wait for kent on this one!

common/core/control/airbrake/airbrake.h
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * @brief Simple setter for temperature data (Celsius)
+     * @param float
+     * @return None
+     */
+    void fetchTempData(const float data);
+
+    /**
+     * @brief Updates the state machine of the airbrake based on current state, velocity, and * altitude. Drives the servo motor to deploy and retract airbrakes. 
+     *
```
pasinponsww, line 49: clangd might messed it up lol checkout why

common/core/control/airbrake/airbrake.h
```diff
@@ -0,0 +1,88 @@
...
+     * 
+     * @param void
+     * @return None
+     */
+    void update(void);
+
+private:
+    AirbrakeState state = AirbrakeState::PRELAUNCH;
+
+    // It appears that temperature is not used (yet).
```
pasinponsww: We need to include **bmp390.h** to integrate the pressure and temperature sensor. To handle the telemetry, we should implement a helper function in utils to dynamically calculate both the sea-level baseline altitude and real-time altitude changes. Additionally, we should log the temperature data to send back to the ground station, which can be incorporated into an upcoming sensor_health() function

(I think i did mention this in the utils folder but I'll said so again here)

Image

common/core/control/airbrake/utils/airbrake_utils.h
```diff
@@ -0,0 +1,60 @@
...
+ *       g   = gravity acceleration constant (m / s^2)
+ *       M   = Molar mass for dry air (kg / mol)
+ *       
+ * @param init_temp The initial temperature read by the barometer (in celsius)
+ * @param init_press The initial pressure read by the barometer
+ * @param curr_press The current pressure reading by the barometer
+ * @return The calculated altitude in meters
+ */
+float calc_altitude(float init_temp, float init_press, float curr_press);
+
```
pasinponsww, line 35: Forgot to mention to calculate altitude. Bmp390 will give you both (get_pressure(), get_temperature() so you just need to Bmp390& reference to where the source of the 'press' comes from !

app/airbrake_test/main.cc
```diff
@@ -0,0 +1,37 @@
...
+
+int main(int argc, char* argv[])
+{
+    bsp_init();
+    Board hw = get_board();
+
+    // should probably pass this in the airbrake ye?
+    monitor_imu(hw.imu);
+    monitor_imu(hw.bmp390);
+
```
pasinponsww, line 17: Once we proof that the airbrake_app works. We going to have a cli app for this (airbrake pcb board) board specifically. Where we don't have to write 2 lines of code, we going to wrap it so it's cleaner as an interface

so airbrake logic + telemetry (sending data to telemetry board logic -- comprise of monitor_imu, monitor_baro, STATE of airbrake, STATE of etc...) into one single cli 

Also you define monitor_imu 2 times, switch monitor_imu(hw.bmp390); -> monitor_baro ...

app/airbrake_test/main.cc
```diff
@@ -0,0 +1,37 @@
...
+    press = hw.bmp390.get_pressure();
+
+    Airbrake airbrake(motor);
+
+    while (1)
+    {
+        hw.imu.read_all(data);
+        airbrake.fetchImuData(data);
```
pasinponsww: no caMel case ngl, snake case plz :)

pasinponsww (changes requested):
Looks better but some of the stuff are need to be fix!

common/core/control/airbrake/utils/airbrake_utils.h
ioseph01: I will combine the two functions.

common/core/control/airbrake/airbrake.h
ioseph01: `temperature` not used yet but `init_temperature` used in the altitude formula.

ioseph01: Real time altitude change would be: calculated current altitude - last measured altitude) / time ?
No idea how to calculate sea-level baseline altitude.

pasinponsww: I send the document on the sea level formula by NASA, just take a look and maybe thinking about at the boot up of the hardware, how it should be initialized (we can talk again when Anton got some resolution)

ioseph01: Was thinking of maybe setting base measurements like base altitude, temperature while state at prelaunch. Idk

pasinponsww: It should have a threshold checkout the physic of this world, some baseline airpressure of November not too hot not too windy LOLLLL

common/core/control/airbrake/utils/airbrake_utils.h
pasinponsww: ALright nice

pasinponsww (commented):
Obviously, you going to need to refactor some stuff but based on the context and new changes this is fine for now. New information today tonight

common/core/manager/sensor_data.h
```diff
@@ -0,0 +1,24 @@
...
+{
+
+/**
+ * @brief Container for sensor data
+ */
+struct SensorData
+{
+    Bno055Data imu;
+    float pressure;
+    float temperature;
```
pasinponsww, line 21: ngl though maybe we can add a struct of the pressure and temp just so that we can have a better cleaner references lol but it doesn't matter for now.

common/core/manager/h7/st_sensor_manager.h
```diff
@@ -0,0 +1,55 @@
...
+ * @brief Sensors that can be passed in managed by the Sensor Manager.
+ * Ptr's because they may be optional or not exist.
+ */
+struct StSensorMgrParams
+{
+    Bno055* const bno055;
+    Bmp390* const bmp390;
+};
+
+class HwSensorMgr : SensorMgr
```
pasinponsww: public SensorMgr

common/core/manager/h7/st_sensor_manager.cc
```diff
@@ -0,0 +1,45 @@
...
+{
+}
+
+SensorData HwSensorMgr::get_latest_sensors() const
+{
+    return data;
+}
+bool HwSensorMgr::update_sensor_data()
+{
+    /* Check if barometer exists and fetch data */
```
pasinponsww, line 32: Should add **nullptr** logic assuming that if this sensor does not exist on the custom hardware, it should exit? Think about every board that goes through this sensor manager; it should be able to point to the sensors that actually exist. If a sensor does not exist, discard it so that the hardware level knows.

pasinponsww (changes requested):
LGTM but a little bit of hot-fix (if you counter-arg lmk if their better solution lol)

common/core/manager/h7/st_sensor_manager.cc
ioseph01: You want it just to exit prematurely? I was thinking that users only past in what sensors they want in the init params. Whatever sensors do exist, update corresponding data. It checks if it does exist or not with the nullptr check `if (sensors.sensor)`

common/core/manager/sensor_data.h
ioseph01: Hotfix for bmp for the future?

pasinponsww: nvm lol

ioseph01:
Temperature data from barometer is no longer used if using the standard sea-level temperature 288.15 K

ioseph01:
Sensor manager not hooked up in `main.cc` yet

ioseph01:
Maybe an init function that gets a lot of barometer readings for pressure to get the average for initial pressure.

common/core/control/airbrake/utils/airbrake_utils.cc
```diff
@@ -0,0 +1,48 @@
...
+{
+    float exp = kUnivGasConst * kTempLapseRate / kGravity / kAirMolarMass;
+    return kSeaLevelTemp / kTempLapseRate *
+           (1 - std::pow(curr_press / init_press, exp));
+}
+
+float calc_velocity(float v0, float aZ, float time)
+{
+    return v0 + aZ * time;
+}
```
pasinponsww, line 21: maybe split into math/ utils/ some function could be pulling out though can be reuse in the future as well not coupled to airbrake only. Could be camera gantry.

pasinponsww (changes requested):
A little change

ioseph01:
Currently waiting for:
1. Servo motor
2. Extended Kalman Filter
3. MPC

PR #124, opened 2026-05-06
jc-crc

### LINKED ISSUE #120: CRC Driver
Data Framing 

[ID][LENGTH][PAYLOAD][CRC]

ioseph01:


common/drivers/checksum/crc.h
```diff
@@ -0,0 +1,30 @@
...
+public:
+    /**
+   * @brief Compute checksum from data
+   * 
+   * @param data block of memory to compute on
+   * @return true if successful, false otherwise
+   */
+    virtual bool compute(std::span<uint8_t> data) = 0;
+    // virtual bool compute(std::span<uint16_t> data) = 0;
+    // Idk what sizes we use yet, I do uint8_t for now
```
pasinponsww: Make it scalable so you can pass a 7, uint8_t size, 16, 32, etc. 

Image

common/drivers/platform/stm32h7/st_crc.h
```diff
@@ -0,0 +1,81 @@
...
+    WORD
+};
+
+enum class CrcPolySize : uint8_t
+{
+    BIT_SIZE_32 = 0,
+    BIT_SIZE_16,
+    BIT_SIZE_8,
+    BIT_SIZE_7
+};
```
pasinponsww, line 46: Don't forgot doc strings! and their are a different mode of operation assuming that make it scalable where the users be able to switching between modes as well polynomial, and normal CRC fixed size.

common/drivers/platform/stm32h7/st_crc.cc
```diff
@@ -0,0 +1,58 @@
...
+    : base_addr{params_.base_addr},
+      settings{params_.settings},
+      inital_crc{params_.initial_crc},
+      generator_polynomial{params_.generator_polynomial}
+{
+}
+
+bool HwCrc::init()
+{
+    // REV OUT, REV IN, POLYSIZE
```
pasinponsww, line 45: Add edge cases handling for timer too apparently it's only operate at 32-bit AHB bus only

Image

pasinponsww (changes requested):
Some a little bit changes need to be make!

ioseph01:
Screenshot 2026-05-15 182103

pasinponsww (approved):
LGTM! just need to test it on the actual hardware!

ioseph01:
Max breakpoints for me while debugging was like 6 so be careful of that when testing it out.
Screenshot 2026-06-05 162656
Screenshot 2026-06-05 162713
Screenshot 2026-06-05 162733

ioseph01:
Test case 2
Screenshot 2026-06-05 170334

Screenshot 2026-06-05 170322
Screenshot 2026-06-05 170243

app/crc_test/main.cc
```diff
@@ -0,0 +1,59 @@
...
+        hw.crc.compare(std::span(data16_2), kActualCrc_2);
+        for (volatile uint32_t i = 0; i < 100000; i++);
+
+        /* uint32_t test */
+
+        hw.crc.compare(std::span(data32_1), kActualCrc_1);
+        for (volatile uint32_t i = 0; i < 100000; i++);
+
+        hw.crc.compare(std::span(data32_2), kActualCrc_2);
+        for (volatile uint32_t i = 0; i < 100000; i++);
```
pasinponsww: So this is working on theory of the software level do you think you can test this with the mock uart where you have the tx and feed it back to your own nucleo board in the rx. 

you can absolutely simulate a CRC error on a UART line by physically shaking or disconnecting the jumper wire while data is actively transmitting. So you can see how the noise reacts or pick up during the transmission.

And you can monitor through passing it in the serial term (write uart to the comm) or YAT monitor!

pasinponsww (changes requested):
A bit more testing and then we can merge!

ioseph01:
https://github.com/user-attachments/assets/51a51a66-76e6-4848-b317-4b8c3ac9d581

pasinponsww (approved):
LGTM! the current test is fine but once we have a chance to integrate that over to the usart-pipe we can defo test how it behaves at high-level.

PR #127, opened 2026-06-03
kh-sysclk

kentnhong:
Configurable clock tree for clock sources HSI64 MHz and HSE8 MHz. Can configure SYSCLK, CPU, HCLK, PCLK1, PCLK2, PCLK3, PCLK4 frequency as anything as long as they are under the maximum allowed frequency for their respective bus.


common/drivers/platform/stm32h7/st_sysclk.h
```diff
@@ -0,0 +1,165 @@
...
+    DIV1 = RCC_HCLK_DIV1,
+    DIV2 = RCC_HCLK_DIV2,
+    DIV4 = RCC_HCLK_DIV4,
+    DIV8 = RCC_HCLK_DIV8,
+    DIV16 = RCC_HCLK_DIV16,
+    DIV64 = RCC_HCLK_DIV64,
+    DIV128 = RCC_HCLK_DIV128,
+    DIV256 = RCC_HCLK_DIV256,
+    DIV512 = RCC_HCLK_DIV512
+};
```
pasinponsww, line 59: I understand the idea of letting the BSP manually set the prescaler, but this actually makes things harder for the user. They now have to dig through the datasheet and figure out the correct APB prescaler values themselves. It might be better to include a README template that lists all valid prescaler combinations along with a few practical examples.

pasinponsww (dismissed):

pasinponsww:
clangd this and add readme and this is done

pasinponsww (approved):
LGTM!

PR #128, opened 2026-06-06
kh-sysclk-hot-fix

pasinponsww:


pasinponsww:
**TODO:**
Fix HCIConfig and USEConfig to support stepping down the frequency. Ensure the D1 Prescaler is tested and the VCO drops accordingly when the frequency changes. Additionally, add operational modes as an option within the object's struct

kentnhong (approved):
Tested on HSI64 and HSE8_BYPASS for multiple frequencies. NOTE: If you use PLL to multiply the clock source frequency, minimum frequency allowed is 64 MHz. LGTM

pasinponsww (commented):
LGTM!

PR #129, opened 2026-06-13
bs-telemetry-pipe

pasinponsww:


app/usart_app/BSP_L476/usart_app_bsp.cc
```diff
@@ -1,3 +1,4 @@
+#include <stdint.h>
```
kentnhong: use c++ standard headers for best practices
```suggestion
#include <cstdint>
```

app/usart_pipe_test/board.h
```diff
@@ -0,0 +1,28 @@
...
+* @brief Board-specific definitions and initialization for the usart_pipe_test application.
+* @version 0.1
+*/
+
+#pragma once
+#include "crc.h"
+#include "gpio.h"
+#include "usart.h"
+
+extern uint8_t rxb;
```
kentnhong: Since we're using C++17 and up, we can now declare global variables as inline instead of extern. This saves us from having to declare a variable header here in board.h and defining it usart_app_bsp.cc. You can just declare it inline and initialize it in the same file and it'll know its a global.
```suggestion
inline uint8_t rxb = 0;
```

app/usart_app/BSP_L476/usart_app_bsp.cc
```diff
@@ -23,6 +24,9 @@ Stml4::HwClock clock{};
 
 StUsart usart(USART2, 4000000, 9600);
 
+// Protobuff rxb buffer
+uint8_t rxb = 0;
```
kentnhong: Remove this and use inline declaration and init to 0 in board.h. This applies to all bsp files by the way, theyll be able to see the same variable from board.h.

app/usart_pipe_test/main.cc
```diff
@@ -0,0 +1,83 @@
...
+ *       Flow: fill PbCmd → pipeline.send() frames + CRC32 → USART TX
+ *             → loopback wire → USART RX → pipeline.receive() validates + decodes → PbCmd
+ */
+
+int main()
+{
+    bsp_init();
+    Board board = get_board();
+
+    Pipeline pipeline(board.crc);
```
kentnhong: Why is object creation here and not in the bsp?

app/usart_pipe_test/main.cc
```diff
@@ -0,0 +1,83 @@
...
+        tx.msg.imu.linear_accel_z = 0.0f;
+
+        tx.msg.has_baro = true;
+        tx.msg.baro.air_pressure = 101325.0f - (count * 10.0f);
+        tx.msg.baro.altitude = 10.0f + (count * 1.5f);
+
+        pipeline.send(&tx, board.usart);
+
+        // Wait for all bytes to arrive back via loopback (~22ms at 115200 baud)
+        for (volatile uint32_t i = 0; i < 5000000; i++);
```
kentnhong: Use our delay API and not busy-wait empty for loops. Apply everywhere else in this PR.

app/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ add_subdirectory(i2c_test)
 add_subdirectory(bmp390_test)
 add_subdirectory(imu_test)
 add_subdirectory(w25q_test)
-add_subdirectory(usart_app)
```
kentnhong, line 7: Why are we getting rid of this?

kentnhong (changes requested):

common/core/protobuff/protobuff.proto
```diff
@@ -0,0 +1,57 @@
...
+    AIRBRAKE_STATUS_UNKNOWN   = 0;
+    AIRBRAKE_STATUS_RETRACTED = 1;
+    AIRBRAKE_STATUS_DEPLOYED  = 2;
+}
+
+message StatusData {
+    FswStatus      fsw_status      = 1;
+    RocketState    rocket_state    = 2;
+    AirbrakeStatus airbrake_status = 3;
+}
```
kentnhong, line 36: Explicitly define these enums as uint8_t and fix state names when you can.

app/usart_pipe_test/main.cc
pasinponsww: technically I can feed inside the bsp but it's more like a pipe that inherit the hardware specific content. I'll just move it

common/core/protobuff/protobuff.proto
pasinponsww: it's a mock rightnow will replace it later (will fetch from josephs at the end)

app/CMakeLists.txt
pasinponsww: accidently override it

app/usart_app/BSP_L476/usart_app_bsp.cc
```diff
@@ -1,3 +1,4 @@
+#include <cstdint.h>
```
kentnhong: No .h for cstdint

kentnhong (changes requested):

pasinponsww:
on pause right now waiting until @alex finished his rs485! and then we can test! their going to be 2 variant test. The uart TTL test vs the actual rs485 test.

app/usart_pipe_test/main.cc
ioseph01, line 1: When you debug and trace to see the LEN of a frame, can you tell me what the value is?

TY

common/core/network/pipeline/usart_pipe.cc
```diff
@@ -0,0 +1,175 @@
...
+    tx_buffer[3] = static_cast<uint8_t>(kSof);
+    tx_buffer[kSofLen] = static_cast<uint8_t>(payload_len);
+
+    // Compute CRC32 over [SOF][LEN][PAYLOAD]
+    // Note: we do CRC32 b/c the payload is variable length
+    // and we want to avoid copying it into a separate buffer for crc calculation
+
+    uint16_t crc_input_len = kHeaderLen + static_cast<uint16_t>(payload_len);
+    uint32_t crc_val = 0;
+    if (!crc.compute(std::span<const uint8_t>(tx_buffer.data(), crc_input_len),
```
ioseph01, line 52: I see crc.compute in send which is good. Is crc.compare for receive going to be used somewhere or is that handled by someone else?

common/core/protobuff/protobuff.proto
ioseph01, line 1: More a question on protobuffers for my own info but do they change byte size if some fields are unused or unfilled? If so, I assume all fields will be used so frames are all the same size unless we want to save space in w25q.

ioseph01: Also, I dont exactly know the byte size diff between .proto and .cc types

common/core/protobuff/protobuff.proto
```diff
@@ -0,0 +1,57 @@
...
+    ROCKET_STATE_UNKNOWN        = 0;
+    ROCKET_STATE_IDLE           = 1;
+    ROCKET_STATE_POWERED_FLIGHT = 2;
+    ROCKET_STATE_COAST          = 3;
+    ROCKET_STATE_APOGEE         = 4;
+    ROCKET_STATE_DESCENT        = 5;
+    ROCKET_STATE_LANDED         = 6;
+}
+
+enum AirbrakeStatus {
```
ioseph01, line 22: This will be updated to match something like
```cc
enum class AirbrakeState : uint8_t
{
    PRELAUNCH = 0,             // Not launched yet
    LAUNCHED,                  // 6 second interval from launch
    AIRBRAKES_DEPLOYING,       // Deploy airbrakes after 6 seconds
    AIRBRAKES_FULLY_DEPLOYED,  // When at 30 degrees; not sure when to know though
    AIRBRAKES_RETRACTING,      // Should do before target apogee
    AIRBRAKES_FULLY_RETRACTED,  // Again, dont know how to know if retracted
    RECOVERY_POPPED,            // Job finished
};
```
later on.

common/core/protobuff/protobuff.proto
ioseph01, line 1: I have a sensor manager class in airbrake branch. It uses this SensorData struct currently.
```cc
/**
 * @brief Container for sensor data
 */
struct SensorData
{
    Bno055Data imu;
    float pressure;
    float temperature;
};
```
Let me know how we can match them.

common/core/network/pipeline/usart_pipe.cc
pasinponsww: This should be handle at the ground software instead! for reason is that having it sit in the telemetry app won't do much though the hardware rs485 bus line  could cause a noise and we can discard early at the telemetry board but we found it redundant to discard on the air. We instead crc.compare() at the ground which is where validation and discard all noisy/mess-up data that pass down from loRa

common/core/protobuff/protobuff.proto
pasinponsww: Thank you for letting me know. I'm assuming you have this in the airbrake branch correct me if I'm wrong. I'll just feed it in here and make it encode and encasulate what you have in payload!

common/core/protobuff/protobuff.proto
pasinponsww: Sounds good. But yeah it's not finalized so this is a mock currently.

pasinponsww:
@note to myself
- need to fix the field inside the protobuf

app/usart_pipe_test/main.cc
pasinponsww: I never check the LEN of a 'total frame' 

I only know the payload length only which is live in 
   uint16_t frame_len = kFrameOverhead + static_cast<uint16_t>(payload_len);
but their is other define content that adds up all of the frame together which is live inside the private class

```
 // Format of the frame: [SOF:4][LEN:1][PAYLOAD:N][CRC32:4][EOF:4]
    static constexpr uint32_t kSof{0xAB6B0BAA};
    static constexpr uint32_t kEof{0x67676767};
    static constexpr uint8_t kSofLen{4};
    static constexpr uint8_t kHeaderLen{kSofLen + 1};  // SOF + LEN byte
    static constexpr uint8_t kCrcLen{4};
    static constexpr uint8_t kEofLen{4};
    static constexpr uint16_t kBufSize{256};
    static constexpr uint8_t kFrameOverhead = kHeaderLen + kCrcLen + kEofLen;
    static constexpr uint16_t kMaxPayloadLen = kBufSize - kFrameOverhead;
```

I think it's good that you point out I will make a getter function that capture the total legnth per LEN of a frame

app/usart_pipe_test/BSP_H723/h723_board.cc
```diff
@@ -0,0 +1,126 @@
...
+
+    // Interrupt-driven RX: the USART3 ISR grabs each incoming byte into the
+    // pipeline's ring buffer, so receive() just decodes whole frames and never polls.
+    NVIC_SetPriority(USART3_IRQn, 0);
+    NVIC_EnableIRQ(USART3_IRQn);
+
+    return ret;
+}
+
+/* I will make a helper and throw this in utils (delay.h) */
```
ioseph01: Did you do it?

common/core/network/pipeline/usart_pipe.cc
```diff
@@ -0,0 +1,194 @@
...
+
+    // Build header — SOF big-endian, then LEN byte
+    tx_buffer[0] = static_cast<uint8_t>(kSof >> 24);
+    tx_buffer[1] = static_cast<uint8_t>(kSof >> 16);
+    tx_buffer[2] = static_cast<uint8_t>(kSof >> 8);
+    tx_buffer[3] = static_cast<uint8_t>(kSof);
+    tx_buffer[kSofLen] = static_cast<uint8_t>(payload_len);
+
+    // Compute CRC32 over [SOF][LEN][PAYLOAD]
+    // Note: we do CRC32 b/c the payload is variable length
```
ioseph01: Why is the payload variable length? Are some fields going to be unused? Idk much about protobuffers

common/core/protobuff/docs/sensor_message_guide.md
```diff
@@ -0,0 +1,114 @@
...
+class PbCmd
+{
+public:
+    RocketMessage msg;   // <- this already is your "sensor data struct"
+    int encode(uint8_t* buffer, size_t maxlen) const;
+    bool decode(const uint8_t* buffer, size_t len);
+};
+```
+
+Don't add a parallel `SensorData` struct that mirrors `RocketMessage` and
```
ioseph01, line 24: So who should be defining the RocketMessage struct? Telemetry or airbrake app?

common/core/protobuff/protobuff.proto
```diff
@@ -0,0 +1,68 @@
...
+    float quat_w         = 10;
+    float linear_accel_x = 11;
+    float linear_accel_y = 12;
+    float linear_accel_z = 13;
+    float gravity_x      = 14;
+    float gravity_y      = 15;
+    float gravity_z      = 16;
+}
+
+message BaroData {
```
ioseph01, line 57: Currently, the barometer has a function to return temperature and return air pressure. There is no BaroData struct as far as I know. Would I have to make one that would package the air pressure, temperature, and a calculated altitude?

common/core/protobuff/docs/sensor_message_guide.md
kentnhong: Depends. If RocketMessage means the payload + framer stuff, then telem. If its just the payload data then it should be airbrake. I'm assuming RocketMessage is just the payload because from my understanding protobuff handles the payload layer and then our telem pipeline slaps headers and footers after.

common/core/protobuff/protobuff.proto
kentnhong: I don't think thats necessary tbh. We were just planning to have baro, imu, and rocket state in one airbrake data struct. However, if you think its worth organizing each entry as a struct so we know what variable is part of what entry, then I think its worth. We could do one struct and have 3 structs inside that struct for organization. Just depends how protobuff handles nested structs.

common/core/network/pipeline/usart_pipe.cc
kentnhong: The payload is not going to be variable length. I think Bex is just saying CRC32 handles variable length data.

common/core/network/pipeline/usart_pipe.cc
```diff
@@ -0,0 +1,214 @@
...
+    if (msg == nullptr)
+        return false;
+
+    /* NEW: Format of the frame: [SOF:4][LEN:1][PAYLOAD:kMaxPayloadLen, zero-padded][CRC32:4][EOF:4] */
+    /* OLD: Format of the frame: [SOF:4][LEN:1][PAYLOAD:N][CRC32:4][EOF:4] */
+
+    // Encode protobuf payload after the header slot
+    int payload_len =
+        msg->encode(tx_buffer.data() + kHeaderLen, kMaxPayloadLen);
+    if (payload_len <= 0)
```
kentnhong: Our payload length is always fixed so I think we shouldnt allow anything to go through thats not of size kMaxPayloadLen. Also kMaxPayloadLen should just be kPayloadLen because saying its max size is confusing like it can vary.

ALSO in the future when we have payload length that varies, we should allow anything between 0 and kMaxPayloadLen. 0 payload length is usually used for quick pings back and forth between ground and fsw.
```suggestion
    if (payload_len != kPayloadLen)
    {
       // Replace payload msg with error msg of same size and send that instead
       // Or you can just return false up to you
    }
```

common/core/network/pipeline/usart_pipe.cc
```diff
@@ -0,0 +1,214 @@
...
+    tx_buffer[crc_offset + 0] = static_cast<uint8_t>(crc_val);
+    tx_buffer[crc_offset + 1] = static_cast<uint8_t>(crc_val >> 8);
+    tx_buffer[crc_offset + 2] = static_cast<uint8_t>(crc_val >> 16);
+    tx_buffer[crc_offset + 3] = static_cast<uint8_t>(crc_val >> 24);
+
+    // Append EOF big-endian
+    tx_buffer[crc_offset + kCrcLen + 0] = static_cast<uint8_t>(kEof >> 24);
+    tx_buffer[crc_offset + kCrcLen + 1] = static_cast<uint8_t>(kEof >> 16);
+    tx_buffer[crc_offset + kCrcLen + 2] = static_cast<uint8_t>(kEof >> 8);
+    tx_buffer[crc_offset + kCrcLen + 3] = static_cast<uint8_t>(kEof);
```
kentnhong, line 70: Why are we using 2 different endians? Wouldn't it be easier if CRC32 was also Big Endian? Makes parsing data a lot easier so we don't have to keep track of what we have to swap.

common/core/network/pipeline/usart_pipe.h
```diff
@@ -0,0 +1,107 @@
...
+    * @param usart Reference to the Usart instance to poll.
+    */
+    void poll_usart(Usart& usart);
+
+    /**
+    * @brief Process the RX buffer to extract and validate a complete frame, then decode the protobuf message.
+    * @param msg Pointer to the PbCmd message to populate with the decoded data.
+    * @return True if a valid frame was processed and decoded, false otherwise.
+    */
+    bool process_frame(PbCmd* msg);
```
kentnhong, line 117: This is unused right? Since we discussed that we would just be forwarding the entire frame to ground through LoRa and then processing and validating on ground. I think we would still need a function to read a frame  from rx ringbuf and return its contents so that main can forward it to SX1262 driver. So maybe a getter for the rx buf so main can pass it to SX1262?

common/core/protobuff/protobuff.proto
```diff
@@ -0,0 +1,68 @@
+syntax = "proto3";
+
+// TODO: This is not finalized (Field could be changed)
+// According to AirBrakeApp & LoRa transciever maximum capabilities
+
+enum FswStatus {
+    FSW_STATUS_UNKNOWN = 0;
+    FSW_STATUS_IDLE    = 1;
+    FSW_STATUS_ACTIVE  = 2;
+}
```
kentnhong, line 10: Instead of IDLE and ACTIVE, we should have NOMINAL and ERROR. Don't think our fsw is ever going to be IDLE haha.
```suggestion
enum FswStatus {
    UNKNOWN = 0;
    NOMINAL = 1;
    ERROR = 2;
}
```

common/core/protobuff/protobuff.proto
```diff
@@ -0,0 +1,68 @@
...
+
+enum AirbrakeStatus {
+    AIRBRAKE_STATUS_PRELAUNCH                = 0;  // Not launched yet
+    AIRBRAKE_STATUS_LAUNCHED                 = 1;  // 6 second interval from launch
+    AIRBRAKE_STATUS_AIRBRAKES_DEPLOYING      = 2;  // Deploy airbrakes after 6 seconds
+    AIRBRAKE_STATUS_AIRBRAKES_FULLY_DEPLOYED = 3;  // When at 30 degrees; not sure when to know though
+    AIRBRAKE_STATUS_AIRBRAKES_RETRACTING     = 4;  // Should do before target apogee
+    AIRBRAKE_STATUS_AIRBRAKES_FULLY_RETRACTED = 5;  // Again, dont know how to know if retracted
+    AIRBRAKE_STATUS_RECOVERY_POPPED          = 6;  // Job finished
+}
```
kentnhong, line 30: Let's discuss this with the rest of the leads and see what they think is useful for these states.

kentnhong (changes requested):

PR #130, opened 2026-06-14
mc-sitl

mariocruz9456:
SITL in the works, airbrakes section completed (I believe)


app/sitl_test/config/airbrake.json
```diff
@@ -0,0 +1,19 @@
+{
```
pasinponsww, line 1: so this is good but the coeffient and vars will have more to feed into this sitl, I will update you regarding that

app/sitl_test/apps/airbrake_app.cpp
```diff
@@ -0,0 +1,87 @@
+#pragma "once"
```
pasinponsww: I know you name your file ".cpp" change it to ".cc" please just so it's consistent 

https://stackoverflow.com/questions/18590135/what-is-the-difference-between-cc-and-cpp-file-suffix

app/sitl_test/apps/airbrake_app.cpp
```diff
@@ -0,0 +1,87 @@
...
+        // Return current airbrake command
+        AirbrakeState getCommand() const {
+            return state;
+        }
+    
+        private:
+            double deployment = 0.0;
+            double target_apogee_m = 3000.0;
+            double deploy_altitude_m = 800.0;
+            double kp = 0.001;
```
pasinponsww: it is convention to initialize your var in the private in 
```
double deployment {0.0};
double target_apogee_m{3000.0};
double deploy_altitude_m{800.0};
double kp{0.001}
```

app/sitl_test/apps/airbrake_app.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+                double predicted_apogee = data.altitude_m + (data.velocity_mps * data.velocity_mps) / (2.0 * g);
+
+                double error = predicted_apogee - target_apogee_m;
+                double output = kp * error;
+
+                if (output < 0.0) output = 0.0;
+                if (output > 1.0) output = 1.0;
+                return output;
+            }
```
pasinponsww: This is okay but I might need more context of feeding different equation and condition commend up top if you think it is necessary for user to understand context.

** Once Joseph get new updates on the airbrakeApp. You have to make some changes !

app/sitl_test/main.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+                AirbrakeState airbrake = AirbrakeApp::getCommand();
+
+                double brake_drag_accel = -airbrake.deployment * 20.0;
+                acceleration_mps2 = -9.81 + brake_drag_accel;
+                velocity_mps += acceleration_mps2 * time_step_s;
+                altitude_m += velocity_mps * time_step_s;
+                time_s += time_step_s;
+            }
+            cout << "Simulation ended at time: " << time_s << " seconds." << endl;
```
pasinponsww, line 137: I might need a demo from you when you feeding in the fake data. And show this at the debut output level. I want to see both failed case, success case, stale case, noisy case, and many other scenario you can test playing around what you feed inside the json.

pasinponsww (changes requested):
Looks good to me but need some explanation and wait on joseph finalized airbrakeApp.

mariocruz9456:
[https://drive.google.com/file/d/1GuWoo59KIyOEqF8WuSzbB4wfR61d_WxK/view?usp=sharing](url)
video demo of airbrakes system

app/sitl_test/sim/rocket_state.h
```diff
@@ -0,0 +1,14 @@
...
+{
+    PRELAUNCH,
+    LAUNCHED,
+    BRAKES_DEPLOYING,
+    BRAKES_DEPLOYED,
+    BRAKES_RETRACTING,
+    BRAKES_RETRACTED,
+    RECOVERY_POPPED,
+    LANDED
+};
```
pasinponsww: define in the uint8_t

Like this: 
enum class RocketState : uint8_t
{
}

Though you're not flashing anything from the stil into the avionics but it guarantees a deterministic 1-byte memory footprint, prevents undefined behavior during out-of-range casts, and enables safe cross-platform serialization. It's usually uint32_t by default which takes up the compute space.

app/sitl_test/sim/recovery_command.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+// Represents the state of the recovery system (not deployed or deployed)
+struct RecoveryState
+{
+    bool deployment = false;
+    // false = not deployed
+    // true = deployed
+};
```
pasinponsww, line 9: I get that this is the mock recovery but assume different deployment scenario you can think of, all instance that would cause it or logic that send it to recovery state.

app/sitl_test/apps/airbrake_app.cpp
pasinponsww: Please tackle this one please ! or maybe delete this file I'm not sure what it's look like on your branch please make a hot-fix!

app/sitl_test/apps/airbrake_app.cc
```diff
@@ -0,0 +1,200 @@
+#pragma once
+
```
pasinponsww, line 2: FYI, if you're looking at the Airbrake app, there will likely be a way for us to normalize the structure for feeding in fake data.

On the real rocket, we reference all of the sources BNO055, BMP390, and the servo (technically not a sensor, but still an external source) and feed them into a single hub. That hub is then passed down through the config as parameters.

We might want to do something similar here and normalize everything so your simulations can feed fake data through a single struct. That would make the interface look much cleaner and easier to work with.

This is mostly a structural consideration, and I'm just giving you some context on what might change in the real `airbrake_app`. You can ask Joseph about it, or wait until he gets to the `HwSensorManager` abstraction layer. He should be able to provide the format of his struct so we can normalize everything consistently.

pasinponsww (changes requested):
Lookds good to me but need some adjustment and structure (based on the gc conversation) please contact joseph for more context !

PR #132, opened 2026-06-20
ap-rs485

alexanderpulido:


common/core/periph/rs485/thvd1420.h
```diff
@@ -0,0 +1,61 @@
+/**
+ * @brief THVD1420 RS485 Transceiver Driver using STM32H7 Hardware DE.
+ */
+
+#pragma once
+
+#include "stm32h723xx.h"
```
pasinponsww: I have question, why do you have the stm32h723xx.h instance in here? 

Because usually when interfacing with sensor, transceiver, motors, etc. You pass down the object of Usart as a tool for you to use (Since this peripheral is there for you).

common/core/periph/rs485/thvd1420.h
```diff
@@ -0,0 +1,61 @@
...
+{
+namespace Stmh7
+{
+
+/**
+ * @brief Configuration parameter block for initializing the RS485 peripheral.
+ */
+struct RS485Params
+{
+    USART_TypeDef*
```
pasinponsww: I guess this already comes with the USART instance, but could you explain why you need it again when `st_usart.h/.cc` already handles all of the UART-specific functionality?

From my understanding, this driver is only responsible for controlling the THVD transceiver's direction (TX/RX). It simply switches the logic that enables transmit or receive mode on the THVD. All UART-related operations such as framing, baud rate configuration, and passing bytes through the USART peripheral are already handled by the existing ST_USART driver.

The THVD itself doesn't care about the contents of the byte stream; it just acts as a physical-layer transceiver. The only logic required here is essentially opening and closing the valve that allows the USART byte stream to pass through in either transmit or receive mode.

common/core/periph/rs485/thvd1420.h
```diff
@@ -0,0 +1,61 @@
...
+#pragma once
+
+#include "stm32h723xx.h"
+#include "usart.h"
+
+namespace LBR
+{
+namespace Stmh7
+{
+
```
pasinponsww: Please adding more comments or @note about what is this tool is and what is needed to operate this tools. 

From my understanding is that it's an external transceiver with logic shift based on the gpio that MCU can sets GPIO_HIGH or GPIO_LOW to be receive/transceiver mode. 

This driver should be easy so it wouldn't take that long !

common/core/periph/rs485/thvd1420.h
```diff
@@ -0,0 +1,61 @@
...
+     * @return true on successful completion of the shift register clear.
+     */
+    bool send(std::span<const uint8_t> txbuf) override;
+
+    /**
+     * @brief Checks the RXNE flag and grabs a byte from the receive register.
+     * @param byte Reference to store the received byte.
+     * @return true if data was ready and captured successfully, false if no data available.
+     */
+    bool receive(uint8_t& byte) override;
```
pasinponsww: Since the USART driver is already fully implemented, you do not need to write any additional driver code. The only requirement is to provide a function that allows the user to switch the transceiver between TX and RX modes. The THVD1420 hardware handles the differential signaling and GPIO HIGH/LOW conversion internally.

common/core/periph/rs485/thvd1420.h
```diff
@@ -0,0 +1,61 @@
...
+
+private:
+    USART_TypeDef* const base_addr;  // Pointer to bare-metal hardware registers
+    uint32_t baud_rate;              // Saved target baud rate
+};
```
pasinponsww: Yeah, overall this THVD shouldn't be complex on your end at all. It's mostly just logic for shifting between high and low states and being able to switch between transmit and receive modes. Of course, you'll want to test the transceiver module. That's where you'll define it in your BSP using the st_usart instance and then define your message in main.cc so you can send it through the transceiver.

common/core/periph/rs485/thvd1420.cc
```diff
@@ -0,0 +1,84 @@
...
+        (0x01
+         << USART_CR1_DEDT_Pos);  // Sample 1-bit period de-assertion buffer
+
+    // Configure for standard 1 Stop Bit
+    base_addr->CR2 &= ~USART_CR2_STOP;
+
+    // Natively automate the THVD1420 Half-Duplex line via hardware
+    base_addr->CR3 |= USART_CR3_DEM;  // Enable Hardware Driver Enable Mode
+    base_addr->CR3 &=
+        ~USART_CR3_DEP;  // Driver Enable Polarity: Active High (DE = 1 to talk)
```
pasinponsww: I see please note the "why" more I need more information.

common/core/periph/rs485/thvd1420.cc
```diff
@@ -0,0 +1,84 @@
...
+    base_addr->CR1 |= (USART_CR1_TE | USART_CR1_RE);
+
+    // RS-485 SPECIFIC: Configure Driver Enable Assertion/De-assertion times
+    // Adds a vital hardware clock-cycle guard band surrounding your bytes
+    base_addr->CR1 &= ~(USART_CR1_DEAT | USART_CR1_DEDT);
+    base_addr->CR1 |=
+        (0x01 << USART_CR1_DEAT_Pos);  // Sample 1-bit period assertion buffer
+    base_addr->CR1 |=
+        (0x01
+         << USART_CR1_DEDT_Pos);  // Sample 1-bit period de-assertion buffer
```
pasinponsww: okay this is make sense to keep if it's mean enable the usart to open the gateway for us to sending data. Can you attach a screenshot or page where it says this?

common/core/periph/rs485/thvd1420.cc
```diff
@@ -0,0 +1,84 @@
...
+    {
+        return false;
+    }
+
+    // Make sure USART is disabled (UE = 0) before doing anything
+    base_addr->CR1 &= ~USART_CR1_UE;
+
+    // Calculate BRR using the live clock speed passed from rs485 test main
+    uint32_t brr_value = peripheral_clock_freq / baud_rate;
+    base_addr->BRR = brr_value;
```
pasinponsww: I guess but please give me more reason why is this different from the mcu stmh7 operation of the usart.

app/rs485_test/main.cc
```diff
@@ -0,0 +1,64 @@
...
+        .clk_src = LBR::Stmh7::Source::HSE8_MHZ_BYPASS,
+        .sysclk = 275'000'000,
+        .d1cpre_psc = LBR::Stmh7::D1cprePrescaler::DIV1,
+        .ahb_psc = LBR::Stmh7::AhbPrescaler::DIV1,    // AHB (HCLK) = 275 MHz
+        .apb1_psc = LBR::Stmh7::Apb1Prescaler::DIV2,  // APB1 Bus = 137.5 MHz
+        .apb2_psc = LBR::Stmh7::Apb2Prescaler::DIV2,
+        .apb3_psc = LBR::Stmh7::Apb3Prescaler::DIV2,
+        .apb4_psc = LBR::Stmh7::Apb4Prescaler::DIV2};
+
+    LBR::Stmh7::HwClock clock_manager(clock_config);
```
pasinponsww: All hard specific please feed into a separate BSP file ! since we don't want to define hardware specific logic in here!

app/rs485_test/main.cc
```diff
@@ -0,0 +1,64 @@
...
+    uint32_t active_usart_clock =
+        static_cast<uint32_t>(clock_manager.get_clock_frequencies().apb1);
+
+    // Configure and allocate the THVD1420 driver instances statically
+    LBR::Stmh7::RS485Params rs485_config{
+        .base_addr = USART3,  // Injects specific base hardware pointer
+        .baud_rate = 115200   // Standard, stable target test rate
+    };
+
+    LBR::Stmh7::Thvd1420 rs485_bus(rs485_config);
```
pasinponsww: This one as well in the BSP but please read other comments regarding the USART usage in this interface.

pasinponsww (changes requested):
Looks okay to me but please answer my question via PR or discord dm! ty!

pasinponsww:
@note The THVD1420 is a purely physical-layer device. It acts like a translator that changes voltage levels, but it does not alter your data, timing, or protocols. This is my understanding.

pasinponsww (dismissed):
LGTM! will merge after Saturday test!

common/core/periph/transceiver/rs485.h
```diff
@@ -0,0 +1,74 @@
...
+{
+class Rs485
+{
+public:
+    enum class Direction
+    {
+        RECEIVE = 0,  /**< DE=0 (Disable TX), RE=0 (Enable RX). */
+        TRANSMIT = 1, /**< DE=1 (Enable TX), RE=1 (Disable RX). */
+        LOOPBACK = 2  /**< DE=1 (Enable TX), RE=0 (Enable RX simultaneously). */
+    };
```
pasinponsww: putting in enum is fine but would be better if you do ->

`enum class Direction : uint8_t `

because guarantees the enum takes exactly 1 byte of memory, whereas a normal un-scoped enum without a defined type defaults to an int (typically 4 bytes).

common/core/periph/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 add_subdirectory(chip_select)
 add_subdirectory(bmp390)
 add_subdirectory(flash_chip)
-add_subdirectory(bno055)
\ No newline at end of file
+add_subdirectory(bno055)
+add_subdirectory(transceiver)
```
pasinponsww: so ngl maybe just rename the transceiver folder name to thvd1420 

so we know what the component it is!

app/rs485_test/board.h
```diff
@@ -0,0 +1,40 @@
...
+ * @brief Architecture wrapper aggregating system dependencies for the hardware test rig.
+ */
+struct Board
+{
+    Usart& usart; /**< Reference to bare-metal serial controller. */
+    Gpio& rx;     /**< Reference to underlying UART RX pin hardware resource. */
+    Gpio& tx;     /**< Reference to underlying UART TX pin hardware resource. */
+    Rs485&
+        rs485; /**< Reference to the platform driver handling independent DE and /RE lines. */
+};
```
pasinponsww, line 26: not sure if clangd mess it up please check

pasinponsww (changes requested):
A little bit of changes here! good attempt!

app/rs485_test/main.cc
```diff
@@ -0,0 +1,46 @@
...
+    Board& board = get_board();
+    board.rs485.init();
+
+    while (1)
+    {
+        board.rs485.set_direction(
+            Rs485::Direction::LOOPBACK);  // PA4 goes HIGH, PA5 goes LOW
+        board.usart.send(txb);
+
+        for (volatile uint32_t settle_delay = 0; settle_delay < 50000;
```
pasinponsww, line 23: use the delay.h the function that we have! instead of busy-wait delay!

pasinponsww (approved):
For per peripheral yest this is fine! But in the future if it's mean to use the delay at any point. Please use delay.h that we have!

PR #138, opened 2026-07-01
bs-timebase

pasinponsww:


common/drivers/platform/stm32h7/st_timebase.cc
```diff
@@ -0,0 +1,194 @@
...
+bool HwTimebase::set_period(std::chrono::microseconds period)
+{
+    // Implementation for setting the period of the timebase
+    return true;
+}
+
+void HwTimebase::start()
+{
+    base_addr->CR1 |= TIM_CR1_CEN;
+}
```
kentnhong: Add check to see if timer period and freq were valid and was set before starting.

common/drivers/platform/stm32h7/st_timebase.h
```diff
@@ -0,0 +1,100 @@
...
+    /**
+    * @brief stop the timebase
+    */
+    void stop() override;
+
+    /**
+    * @brief get the current time in microseconds
+    * @return The current time in microseconds
+    */
+    uint64_t now_us() const override;
```
kentnhong: Clarify this is for getting the time since timer was started. Or give the function a better name.

common/drivers/platform/stm32h7/st_timebase.h
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+    * @brief get the elapsed time in microseconds since a given time
+    * @param since The starting time in microseconds
+    * @return The elapsed time in microseconds
+    */
+    uint32_t elapsed_us(uint32_t since) const override;
```
kentnhong: Give better name or example of usage.

common/drivers/platform/stm32h7/st_timebase.h
```diff
@@ -0,0 +1,100 @@
...
+class HwTimebase : public Timebase
+{
+public:
+    explicit HwTimebase(const StTimebaseParams& params);
+
+    /**
+    * @brief init the timebase with a specified tick period
+    * @param tick_period The period of the timebase ticks in microseconds
+    */
+    bool init(std::chrono::microseconds tick_period) override;
```
kentnhong: Don't have to set timer period here. Init can be solely for setting reg values not related to setting timer period and freq.

common/drivers/platform/stm32h7/st_timebase.h
```diff
@@ -0,0 +1,100 @@
...
+    * @return true if the timer frequency was set successfully, false otherwise
+    */
+    bool set_freq(uint32_t new_timer_freq, uint32_t pclk) override;
+
+    /**
+    * @brief Set the period of the desired timebase TIM channel
+    * @param period The desired period in microseconds
+    * @return true if the period was set successfully, false otherwise
+    */
+    bool set_period(std::chrono::microseconds period) override;
```
kentnhong, line 85: Can set a flag when freq and period are valid to indicate that its ok to start timer at the end of the functions. Similarly, reset flag to false in beginning of function so user cant start timer until period and freq are valid. Make sure timer is paused before setting new period and freq.

kentnhong (changes requested):

common/drivers/platform/stm32h7/st_timebase.h
pasinponsww: It's a getter

pasinponsww:
need only to test and we will see if it's working good or not or proof that it works as we setting a desire frequency and time period that we want.

common/drivers/platform/stm32h7/st_timebase.cc
```diff
@@ -0,0 +1,367 @@
...
+        !is_supported_timer(base_addr))
+    {
+        return false;
+    }
+
+    // If interrupts are not enabled, we cannot use the timebase
+    if (!enable_irq)
+    {
+        return false;
+    }
```
kentnhong, line 79: This is fine for now until we need to support using timers without interrupts. In other words, applications that directly check the overflow flag from the timer.

common/drivers/platform/stm32h7/st_timebase.cc
```diff
@@ -0,0 +1,367 @@
...
+
+    // Require an exact integer divider to achieve the desired counter frequency
+    if ((timer_input_hz % kDefaultCounterHz) != 0U)
+    {
+        return false;
+    }
+
+    // Calculate the divider to achieve a 1 MHz counter frequency (1 tick = 1 microsecond)
+    const uint32_t divider = timer_input_hz / kDefaultCounterHz;
```
kentnhong, line 88: This is fine for now, until we want to support more counter frequencies instead of hardcoding each tick to be 1us. Just keep that in mind.

common/drivers/platform/stm32h7/st_timebase.cc
```diff
@@ -0,0 +1,367 @@
...
+        return false;
+    }
+
+    // Disable the timer and its interrupt before configuration
+    base_addr->CR1 &= ~TIM_CR1_CEN;
+    base_addr->DIER &= ~TIM_DIER_UIE;
+
+    // Clear any pending interrupts and disable the NVIC interrupt for the timer
+    NVIC_DisableIRQ(irq);
+    NVIC_ClearPendingIRQ(irq);
```
kentnhong: Idk how I feel about coupling the IRQ controller to the timebase driver since not all timebase drivers necessarily use an IRQ controller. Usually we do all NVIC operations in the bsp file and before driver inits. This is so interrupts dont fire during the inits.

common/drivers/platform/stm32h7/st_timebase.cc
```diff
@@ -0,0 +1,367 @@
...
+    // Configure the timer in upcounting mode with no clock division
+    base_addr->SMCR &= ~TIM_SMCR_SMS;
+    base_addr->CR1 &= ~(TIM_CR1_DIR | TIM_CR1_CMS);
+
+    // Enable auto-reload preload and update request source
+    base_addr->CR1 |= TIM_CR1_ARPE;
+    base_addr->CR1 |= TIM_CR1_URS;
+
+    // Set the prescaler and auto-reload register to achieve the desired counter frequency
+    base_addr->PSC = divider - 1U;
```
kentnhong: Be cautious of implicit conversion. In our case its fine since 1U on an ARM Cortex M7 is a unsigned int 32 bit type. But, if we were on a 64-bit platform, divider - 1U would implicitly get converted into a 64 bit unsigned int.

common/drivers/platform/stm32h7/st_timebase.cc
```diff
@@ -0,0 +1,367 @@
...
+    NVIC_EnableIRQ(irq);
+
+    configured = true;
+    return true;
+}
+
+bool HwTimebase::start()
+{
+    // Check if the timer is configured and the base address is valid
+    if (base_addr == nullptr || !configured)
```
kentnhong, line 128: Also check if timer is already running. We want to make sure its paused before overwriting any values to the regs.

common/drivers/platform/stm32h7/st_timebase.cc
```diff
@@ -0,0 +1,367 @@
...
+
+    // Enable the timer by setting the CEN bit in the CR1 register
+    base_addr->CR1 |= TIM_CR1_CEN;
+    return true;
+}
+
+void HwTimebase::stop()
+{
+    // Check if the timer is configured and the base address is valid
+    if (base_addr != nullptr)
```
kentnhong: Missing check for timer configured? Or is this comment not relevant anymore?

common/drivers/platform/stm32h7/st_timebase.cc
```diff
@@ -0,0 +1,367 @@
...
+        new_counter_freq > timer_clk)
+    {
+        return false;
+    }
+
+    // Check if the timer clock matches the expected input frequency
+    if (timer_clk != timer_input_hz)
+    {
+        return false;
+    }
```
kentnhong: Why add timer_clk as a function param when you can just use timer_input_hz if this value is constant from the time a timebase object is created?

common/drivers/platform/stm32h7/st_timebase.cc
```diff
@@ -0,0 +1,367 @@
...
+    if (divider == 0U || divider > (kMaxPsc + 1U))
+    {
+        return false;
+    }
+
+    // Save the current running state of the timer
+    const bool was_running = (base_addr->CR1 & TIM_CR1_CEN) != 0U;
+
+    // Disable the timer before reconfiguration
+    base_addr->CR1 &= ~TIM_CR1_CEN;
```
kentnhong: Why not use stop() function here?

common/drivers/platform/stm32h7/st_timebase.cc
```diff
@@ -0,0 +1,367 @@
...
+    if (period_ticks > max_period_ticks)
+    {
+        return false;
+    }
+
+    // Save the current running state of the timer before reconfiguration
+    const bool was_running = (base_addr->CR1 & TIM_CR1_CEN) != 0U;
+
+    // Disable the timer before reconfiguration
+    base_addr->CR1 &= ~TIM_CR1_CEN;
```
kentnhong: Same here. No stop() function?

common/drivers/platform/stm32h7/st_timebase.cc
```diff
@@ -0,0 +1,367 @@
...
+     */
+    uint32_t overflow_before;
+    uint32_t overflow_after;
+    uint32_t counter;
+
+    // Loop until a consistent snapshot of the overflow count and counter is obtained
+    do
+    {
+        overflow_before = overflow_count;
+        counter = base_addr->CNT;
```
kentnhong: Does reading from the CNT register change the value of overflow_count? Should add comment if so.

common/drivers/platform/stm32h7/st_timebase.cc
pasinponsww: That's fair

common/drivers/platform/stm32h7/st_timebase.cc
pasinponsww: Thank you for pointing out.

common/drivers/platform/stm32h7/st_timebase.cc
pasinponsww: oops

pasinponsww:
Thank you for pointing out I think I didn't tie things to the define comments good enough yet. Still need to revise it.

common/drivers/platform/stm32h7/st_timebase.cc
pasinponsww: sounds good

common/drivers/platform/stm32h7/st_timebase.cc
pasinponsww: sounds good but I will make it doable for supporting timers w/o interrupts

common/drivers/platform/stm32h7/st_timebase.cc
pasinponsww: done

common/drivers/platform/stm32h7/st_timebase.cc
pasinponsww: that's fair.

common/drivers/platform/stm32h7/st_timebase.cc
pasinponsww: done forgor

common/drivers/platform/stm32h7/st_timebase.cc
pasinponsww: forgor too

common/drivers/platform/stm32h7/st_timebase.cc
```diff
@@ -312,9 +334,21 @@ uint64_t HwTimebase::uptime_us() const
```
kentnhong: I think there might be a fundamental problem with this function still. Interrupt could still fire in between these lines. If that does happen, UIF will be cleared then and the condition on line 329 will never result to true. This means even overflow_after isnt incremented and we'll get the wrong calculation. I would just make the lines reading the timer counter a critical section and disable all interrupts.

Suggestion:

const uint32_t primask = __get_PRIMASK();
__disable_irq();

uint32_t overflow = overflow_count;
uint32_t counter = base_addr->CNT;

if ((base_addr->SR & TIM_SR_UIF) != 0U)
{
    ++overflow;
    counter = base_addr->CNT;
}

if (primask == 0U)
{
   __enable_irq();
}

kentnhong (approved):
lgtm

PR #155, opened 2026-08-28
CMake SITL Build

kentnhong:
Added a SITL target for CMake builds.

image

Can run "./make.sh -t sitl" now.


pasinponsww (approved):
Looks good to me/

PR #157, opened 2026-09-05
jc-inverse-bno055

ioseph01:
This depends on the atmosphere model to output realistic or simulated readings. This is also waiting on the fake-i2c branch.

### DIFF
--- a/sitl/drivers/inverse/hw/inv_bno055.cc
+++ b/sitl/drivers/inverse/hw/inv_bno055.cc
@@ -1,0 +1,137 @@
+#include "inv_bno055.h"
+
+namespace LBR
+{
+
+static bool is_read_only(uint8_t reg_addr)
+{
+    if ((0x08 <= reg_addr && reg_addr <= 0x37) || reg_addr == 0x39 ||
+        reg_addr == 0x3A)
+    {
+        return true;
+    }
+    return false;
+}
+
+InvBno055::InvBno055(uint8_t dev_addr)
+    : dev_addr_{dev_addr}, recent_reg_{Bno055_CHIP_ID_REG}
+{
+}
+
+bool InvBno055::init()
+{
+    register_map_ = {
+        {0x00, 0xA0},  // CHIP ID
+        {0x07, 0x00},  // PAGE_ID
+
+        {0x08, 0x00},  // ACCEL_DATA_X_LSB
+        {0x09, 0x00},  // ACCEL_DATA_X_MSB
+        {0x0A, 0x00},  // ACCEL_DATA_Y_LSB
+        {0x0B, 0x00},  // ACCEL_DATA_Y_MSB
+        {0x0C, 0x00},  // ACCEL_DATA_Z_LS
+        {0x0D, 0x00},  // ACCEL_DATA_Z_MS
+
+        {0x0E, 0x00},  // MAG_DATA_X_LSB
+        {0x0F, 0x00},  // MAG_DATA_X_MSB
+        {0x10, 0x00},  // MAG_DATA_Y_LSB
+        {0x11, 0x00},  // MAG_DATA_Y_MSB
+        {0x12, 0x00},  // MAG_DATA_Z_LSB
+        {0x13, 0x00},  // MAG_DATA_Z_MSB
+
+        {0x14, 0x00},  // GYRO_DATA_X_LSB
+        {0x15, 0x00},  // GYRO_DATA_X_MSB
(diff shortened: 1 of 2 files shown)


ioseph01:
Datasheet for bno055 https://www.mouser.com/datasheet/2/783/BST_BNO055_DS000-1509603.pdf

pasinponsww:
one thing to be aware when writing the imu driver is the init sequence. I hardcoded the mode though a lot of this can be pull off and map to enum. At the moment (back in 2025) we do not care much about scale. So you can mimick one for one with the current driver or make it modular is doable as well as a sitl inverse driver. This is fyi

sitl/drivers/inverse/hw/inv_bno055.cc
```diff
@@ -0,0 +1,137 @@
...
+}
+
+InvBno055::InvBno055(uint8_t dev_addr)
+    : dev_addr_{dev_addr}, recent_reg_{Bno055_CHIP_ID_REG}
+{
+}
+
+bool InvBno055::init()
+{
+    register_map_ = {
```
pasinponsww, line 13: Instead of define all register map like this:

I think you can just add this define field of labels 


...
struct RegisterDef {
    uint8_t address;
    uint8_t reset_value;
    const char* name;
};

static constexpr RegisterDef kRegisters[] = {
    {0x00, 0xA0, "CHIP_ID"},
    {0x07, 0x00, "PAGE_ID"},

    {0x08, 0x00, "ACCEL_DATA_X_LSB"},
    {0x09, 0x00, "ACCEL_DATA_X_MSB"},
    {0x0A, 0x00, "ACCEL_DATA_Y_LSB"},
    {0x0B, 0x00, "ACCEL_DATA_Y_MSB"},
    {0x0C, 0x00, "ACCEL_DATA_Z_LSB"},
    {0x0D, 0x00, "ACCEL_DATA_Z_MSB"},

    {0x0E, 0x00, "MAG_DATA_X_LSB"},
    {0x0F, 0x00, "MAG_DATA_X_MSB"},
    
    // ...
};
...

So you don't have to guess later on. 

Probably better and then just pass it into the init. 

This mean slight reconstruction at the state reading. Have like a variables that is mutable for state and then you can see what it split out rather than mapped it yourselves.

sitl/drivers/inverse/hw/inv_bno055.cc
```diff
@@ -0,0 +1,137 @@
+#include "inv_bno055.h"
+
+namespace LBR
+{
+
+static bool is_read_only(uint8_t reg_addr)
+{
+    if ((0x08 <= reg_addr && reg_addr <= 0x37) || reg_addr == 0x39 ||
```
pasinponsww: I know it's only inverse driver but I think we should define where is this comes from. I think it might better have a different file for all regs table. So we can mapped to state.


This is something I should do in my Imu old driver in all honesty.

sitl/drivers/inverse/hw/inv_bno055.cc
```diff
@@ -0,0 +1,137 @@
...
+        reg_addr == 0x3A)
+    {
+        return true;
+    }
+    return false;
```
pasinponsww: A lot of this though split true or false but mapped it to an enum with all fail or good scenario are better. It can fail like fail read, fail write, etc. 

this is inverse driver (we can technically step thru debug but as a thread grows it's hard to look at it and adding stopping point (we don't even run this on mcu after all it's a runtime on our computer and it logs into serial). 

Return True or false doesn't differentiate much of state it's at. So having enum and also a flag that being trigger for read fault, write fault, opcode fault, etc is easier for diagnosing the pipeline itself. That's how I would do it. And this may applied to inverse i2c as well.

This makes the pipeline sitl become more fault tolerant. And you know what to solve rather than step thru and find where it breaks. 


It's like airbrakeApp what is break what is working, etc. which mapped to status. fairly similar to this as well. 

Though the actual baremetal driver return true or false cuz it's easier to diagnose where it fail and true with OpenOCD toolchain.

pasinponsww (changes requested):
Slightly change to make this fault tolerant and make this more easy to diagnose and state status when you run this sitl pipeline at runtime and logging it

sitl/drivers/inverse/hw/inv_bno055.cc
ioseph01: Changing to vector<RegisterDef>. Rip O(1) lookup

pasinponsww (approved):
I guess we need to test it and see how we can split it in the attach into the log debugging process

PR #162, opened 2026-09-14
jc-bno05-hotfix

ioseph01:
https://www.bosch-sensortec.com/media/boschsensortec/downloads/datasheets/bst-bno055-ds000.pdf


ioseph01:
It looked like the idx was reaching beyond the size of the buf at 37 when allocated was 32. Also the reads mapping to registers looked wrong according to that datasheet.

mariocruz9456 (approved):
looks good

pasinponsww (approved):
YEs this is more safe, this is good that you put it this way

